PR #17, opened 2018-10-10
Java Module: DOCS-2907 update and image

dukesphere:


modules/ROOT/pages/java/java-invoke-method.adoc
```diff
@@ -70,4 +72,26 @@ You can invoke methods using DataWeave functions. This is especially helpful for
 </choice>
 ----
 
-Note that `vars.taxCalculator` contains an instance of `com.foo.TaxCalculator` that is created with `<java:new/>`, as explained above.
\ No newline at end of file
+Note that `vars.taxCalculator` contains an instance of `com.foo.TaxCalculator` that is created with `<java:new/>`, as explained above.
+
+== Compiling Java Code and Imported Java Classes
+
+To compile your custom Java in `src/main/java`, you need to set the `-parameters` compiler flag. For example, in the POM file for your Mule
```
Mrk-Nguyen: So you don't need to set parameters compiler flag in order to use it, we need to say that if you plan to use the actual parameter names in your methods from your custom Java, you need to flip on the parameter flag. Otherwise you need to use arg0, arg1, etc.

Mrk-Nguyen: I suggest we give an example to call a method from com.foo.TaxCalculator using arg0, arg1 and then using the parameter names. We can then tell people that in order to use the parameter names, you need to flip on the parameters flag.

kevintroller (approved):

PR #20, opened 2018-10-16
update spring-security schema version in beans.xml

SunilVu:
For spring-security, dependency version is 4.2.6.RELEASE. But the beans.xml given in the doc references to spring-security-4.1.xsd in schemaLocation. This needs to be changed to spring-security-4.2.xsd to build the project.


george-hoffer:
Thank you for helping improve MuleSoft documentation. This issue was caught and fixed internally.

PR #47, opened 2018-12-04
MULE-16071: Create an HTTP connector overview doc

afelisatti:


george-hoffer (approved):

george-hoffer:
Notes: We no longer use :keyword: statements. Also, headings for procedures should not start with verbs ending in "ing" such as Mapping or Configuring -- use Map or Configure instead.

modules/ROOT/pages/http/http-connector.adoc
```diff
@@ -3,4 +3,155 @@ ifndef::env-site,env-github[]
...
-Release Notes: xref:release-notes::connector/connector-http.adoc[HTTP Connector Release Notes]
\ No newline at end of file
+Release Notes: xref:release-notes::connector/connector-http.adoc[HTTP Connector Release Notes]
+
+The HTTP connector allows declaring HTTP servers that can listen to
+requests and trigger flows, as well as HTTP clients that can communicate with any
+HTTP service. This means that the HTTP connector effectively allows you to both
+expose and consume HTTP based APIs. 
+
+Examples of HTTP based APIs are that the HTTP connector powers both APIkit
```
afelisatti: @george-hoffer, these are not examples of HTTP based APIs. These are the MuleSoft products to create APIs and to consume them.

modules/ROOT/pages/http/http-connector.adoc
```diff
@@ -3,4 +3,155 @@ ifndef::env-site,env-github[]
...
+You can choose specific resource paths to execute a flow, so that when a request
+is received matching that path, Mule runs your flow.
+
+The following example displays either Hello World or Goodbye World depending on how you configure the HTTP connector:
+
+[source,xml,linenums]
+----
+<flow name="hello">
+    <http:listener path="greet" config-ref="serverConfig"/>
+    <set-payload value="Hello World!"/>
```
afelisatti: @george-hoffer, why have you modified the examples? This makes the latter ones nonsense.

afelisatti:
I've fixed the examples and made the client and server sections more consistent with each other. To avoid the back and forth in terms of content I think it's best to use comments on things that you believe are hard to understand or "smell" rather than edit immediately. That way we can also avoid inconsistencies. Let me know what you guys think, @george-hoffer and @dukesphere.

modules/ROOT/pages/http/http-connector.adoc
george-hoffer: I was trying to apply context to the sentence describing these. Do we need to keep this sentence if they aren't directly related to the HTTP connector? ("For example, APIkit and REST Connect both use the HTTP connector internally to work with APIs.")

modules/ROOT/pages/http/http-connector.adoc
george-hoffer: The example that had "excuse" is not a good word choice as it has multiple meanings. "exit" is a clearer action word.

modules/ROOT/pages/http/http-connector.adoc
afelisatti: My idea of this overview is to explain the raw capabilities of the connector and how it maps to what people expect when thinking about HTTP, hence all the server and client talk. Because the subject is intertwined with APIs, which is our main business, I think it's important to note that we actually have larger tooling/products on the area and the HTTP connector is just a part of it.

afelisatti: We should ideally link to docs on those subjects but I have no idea where those are.

dukesphere (approved):
with light copyedit

PR #59, opened 2018-12-21
swap link titles

Mulesoft-Mysti:
The topics had each other's titles in the left Nav (ToC). A tweak to one of the titles to make it clearer what-all it does was also requested.


marianogonzalez (approved):

george-hoffer (approved):
Should get rid of To/About in titles

Mulesoft-Mysti:
Agree, George, but that raises a few issues so I'm saving that for another day. Thanks for the reminder!

PR #72, opened 2019-01-18
Fixed typo for consume and publish-consume operations

dhimate:
Fixed typo for consume and publish-consume operations where <vm:listener> was not visible in HTML view


dhimate:
Fixed typo for consume and publish-consume operations

george-hoffer (approved):

PR #76, opened 2019-01-19
DOCS-3806 note + edits

george-hoffer:
For more info, see https://www.mulesoft.org/jira/browse/DOCS-3806 -- this feedback jira asks for a throws section, which is non-standard for connector guides and is already covered in the File connector reference (https://docs.mulesoft.com/connectors/file/file-documentation). I just added a link to the reference, which is in the same folder as this guide.


wwarner-salesforce:
I ran a local build, and the "See Also" link worked but the one on line 13 was broken. I changed line 13 to match the "See Also" version and rebuilt locally and they both worked, so I think that's the correct syntax. I just made a push to this branch making the same change.

wwarner-salesforce (approved):

PR #85, opened 2019-01-23
DOCS-3459 FTP Copy Move update + DOCS 3458

george-hoffer:
https://www.mulesoft.org/jira/browse/DOCS-3458
https://www.mulesoft.org/jira/browse/DOCS-3459


wwarner-salesforce (approved):
Looks great!

PR #86, opened 2019-01-23
DOCS-3457 section adds on Poll and Match

george-hoffer:
FTP doc update for:
https://www.mulesoft.org/jira/browse/DOCS-3457


george-hoffer:
Closing to handle this from a branch instead of the git UI

PR #92, opened 2019-01-23
MULE-15386: Review/rewrite HTTP listener doc

afelisatti:


modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
 endif::[]
 :keywords: anypoint studio, esb, connectors, http, https, http headers, query parameters, rest, raml
 
-The required minimum settings for the Listener operation are:
+The HTTP listener is a source that allows setting up an HTTP server and triggering flows when HTTP requests are received.
 
-* Host
-* Port
-* Path
+You can choose what methods the source accepts, such as GET, POST or a list with both, and on which path requests will be accepted on, therefore allowing routing the requests through different flows.
```
fsgonz: "or a list of them" instead of a "or a list of both". Otherwise it seems that only GET and POST are supported.

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
 
-If a request matches the path of multiple Listeners, the Listener having the most specific path has precedence, regardless of the order in which the operations are defined. Only one connector handles each request.
-For example, given the following Listener configurations, the second one receives the request when you use this URL: `account/mulesoft/main-contact`
+[TIP]
+===
+To deploy a project to CloudHub, setting the host to *0.0.0.0* is required since it sets up the connection to listen on all interfaces of the machine.
 
-* Path = `account/(...)`
-* Path = `account/(...)/main-contact`
+When you deploy locally for testing purposes, setting the host to *localhost* is better since it means only requests generated inside the server (your local machine) will be received: the app is not vulnerable to external threats.
```
fsgonz: maybe "generated locally: the app is not vulnerable to external threats" instead of "inside the server (your local machine)"

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
+----
+<flow name="http-triggered-flow">
+  <http:listener path="/" config-ref="HTTP_Listener_config"/>
+  <set-payload value="Help! I've been triggered!"/>
+</flow>
+----
 
-In the Path setting, you can use the wildcard asterisk (*) as follows:
+When an HTTP request is matched to a listener, its request data is used to trigger the flow. The body is set as the payload while all other metadata from the request line to the headers are shared as attributes.
+When the triggered flow is done, the HTTP listener uses it's result to generate a proper HTTP response.
```
fsgonz: its result.

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
-Put a wildcard in the middle of a path when you want a request to succeed that has a URL consisting of the prefix and suffix to the asterisk. For example, Path = `account/\*/main-contact`. The HTTP Listener listens for every HTTP request having a URI that starts with `/account/` and ends with `/main-contact`; the segment in the middle can vary. Requests succeed from the following URLs:
+When multiple listeners are defined, requests are always routed to the most specific one. Hence in the previous examples a request with `accountId` `mulesoft` and suffix `main-contact` will be handled by the first listener,
+and any different `accountId` value by the second listener instead of the third.
 
-* `+http://localhost:8081/account/mulesoft/main-contact+`
-* `+http://localhost:8081/account/foo/main-contact+`
+=== About Methods
 
-=== Using a Placeholder in a URI
+Request can also be routed based on the HTTP method received. By default, a listener will handle all methods but these can be restricted. In the following example, `GET`, `POST` and `PUT` requests will be routed to different flows.
```
fsgonz: "Requests can also" or "The request can also"

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
+===
+To deploy a project to CloudHub, setting the host to *0.0.0.0* is required since it sets up the connection to listen on all interfaces of the machine.
 
-* Path = `account/(...)`
-* Path = `account/(...)/main-contact`
+When you deploy locally for testing purposes, setting the host to *localhost* is better since it means only requests generated inside the server (your local machine) will be received: the app is not vulnerable to external threats.
+===
 
-=== Wildcard in Path
+Advanced users can also define whether the connections will be persistent and if not, the timeout they will have.
```
fsgonz: Maybe you shouldn't refer to "Advanced users" but to "Advanced configuration", e.g. "More advanced configuration can be done. For example,"

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
+[source,xml,linenums]
+----
+<flow name="http-triggered-flow">
+  <http:listener path="/" config-ref="HTTP_Listener_config"/>
+  <set-payload value="Help! I've been triggered!"/>
+</flow>
+----
 
-In the Path setting, you can use the wildcard asterisk (*) as follows:
+When an HTTP request is matched to a listener, its request data is used to trigger the flow. The body is set as the payload while all other metadata from the request line to the headers are shared as attributes.
```
fsgonz: maybe not "shared". "propagated" or something like that.

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
-= HTTP Listener Configuration Reference
+= HTTP Listener Reference
 ifndef::env-site,env-github[]
 include::_attributes.adoc[]
 endif::[]
 :keywords: anypoint studio, esb, connectors, http, https, http headers, query parameters, rest, raml
 
-The required minimum settings for the Listener operation are:
+The HTTP listener is a source that allows setting up an HTTP server and triggering flows when HTTP requests are received.
```
dukesphere: "that allows setting up" -> 
"that enables you to set up ... and trigger flows ..."

modules/ROOT/pages/http/http-listener-ref.adoc
dukesphere: "on requests will be accepted on, therefore allowing routing the requests" -> 
"on which to accept requests, thereby allowing the routing of requests"

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
 
-* Host
-* Port
-* Path
+You can choose what methods the source accepts, such as GET, POST or a list with both, and on which path requests will be accepted on, therefore allowing routing the requests through different flows.
 
-You can choose what methods the connector accepts, such as GET or POST. Provide a comma-separated list of the methods that the connector allows in Advanced > Allowed Methods. The default blank list allows all methods. The connector routes incoming requests based on the URL that requests the app.
+Once a request is accepted by the source, the corresponding flow is triggered with the HTTP body as payload and the HTTP metadata as attributes (headers, query params and so on).
```
dukesphere: "source" -> "Event source" or maybe "Mule Event source" 

"query params" -> "query parameters"

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
 
-You can configure the HTTP Listener to accept any request, or to accept only those requests addressed to a specific URI. You set the Path to limit the requests.
+When the flow finishes its execution, the HTTP listener allows customizing the HTTP response based on whether the execution was successful or not, so that different status codes may be returned, for example.
```
dukesphere: "allows customizing" -> "enables you to customize" or "allows for the customization of"

"codes may be" -> "codes can be"  (in tech writing, we typically avoid the word "may" because it can convey the idea of permission to do something, whereas "can" only implies ability to do something)

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
 
-The connector passes the request body to the next element of the Mule flow as the message payload of a Mule message. Headers, query parameters, and URI parameters are passed as attributes. You can access these properties from your Mule application using DataWeave.
+== HTTP Listener Configuration
 
-In response to the request that the Listener receives, you can configure response codes and clarifications for success and error outcomes to present different error messages depending on the outcome.
+To use an HTTP listener, a configuration must be declared with a corresponding connection. This establishes the HTTP server that will listen to requests.
```
dukesphere: "This establishes" -> 
"This declaration establishes"  (even when it's obvious, we're supposed make sure to say what "This" refers  to)

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
 
-== Setting the Host
+Additionally, a base path can be configured that will apply to all listeners using the configuration.
```
dukesphere: "a base path can be configured that will apply to" -> 
"you can configure a base path that applies to all...." 
(we try to use present tense, active voice when possible and so long as we don't turn the sentence into a pretzel)

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
+<http:listener-config name="HTTP_Listener_config" basePath="api/v1">
+  <http:listener-connection host="0.0.0.0" port="8081" />
+</http:listener-config>
+----
 
-When you depoly an app locally, set Host = *localhost* for security. Using localhost listens for incoming requests generated inside the server, so the app is not vulnerable to external threats.
+=== HTTP Listener Connection
 
-Setting Host to `0.0.0.0` sets up the Listener to listen on all interfaces of the machine including the following examples:
+The connection defines the host and port where the server will be set up as well as the protocol to use: HTTP for plain connections or HTTPS for TLS connections.
```
dukesphere: "set up as well" -> "set up, as well as"

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
 
-* `+http://localhost:8081/mybasepath+`
-* `+http://127.0.0.1:8081/mybasepath+`
-* `+http://myhost.some.domain:8081/mybasepath+`
+To allow secure connections through HTTPS, a TLS Context should be defined in the connection as well providing a key store for the server. A trust store can also be provided in case
+two way authentication is desired. See the xref:4.1@mule-runtime/tls-configuration.adoc[TLS Configuration] guide for details.
```
dukesphere: "the xref:4.1@mule-runtime/tls-configuration.adoc[TLS Configuration] guide" -> "xref:4.1@mule-runtime/tls-configuration.adoc[TLS Configuration]"   (without "the" and "guide")

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
-When you depoly an app locally, set Host = *localhost* for security. Using localhost listens for incoming requests generated inside the server, so the app is not vulnerable to external threats.
+=== HTTP Listener Connection
 
-Setting Host to `0.0.0.0` sets up the Listener to listen on all interfaces of the machine including the following examples:
+The connection defines the host and port where the server will be set up as well as the protocol to use: HTTP for plain connections or HTTPS for TLS connections.
 
-* `+http://localhost:8081/mybasepath+`
-* `+http://127.0.0.1:8081/mybasepath+`
-* `+http://myhost.some.domain:8081/mybasepath+`
+To allow secure connections through HTTPS, a TLS Context should be defined in the connection as well providing a key store for the server. A trust store can also be provided in case
```
dukesphere: "a TLS Context should be defined" -> "you should define a TLS context" or "you need to define a TLS context" (present, active)

"connection as well as providing a" -> "connection and provide a"

"A trust store can also be provided in case two way authentication is desired." -> 
"If you need two-way authentication, you can also provide a trust store." 
or maybe
"You can also provide a trust store if you need two-way authentication."

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
+    <tls:trust-store path="keystore.jks" password="MyP455W0rD"/>
+  </tls:context>
+</http:listener-connection>
+----
 
-If a request matches the path of multiple Listeners, the Listener having the most specific path has precedence, regardless of the order in which the operations are defined. Only one connector handles each request.
-For example, given the following Listener configurations, the second one receives the request when you use this URL: `account/mulesoft/main-contact`
+[TIP]
+===
+To deploy a project to CloudHub, setting the host to *0.0.0.0* is required since it sets up the connection to listen on all interfaces of the machine.
```
dukesphere: "since" -> "because"
(we usually avoid "since" because it can signify temporality or causality, whereas "because" is just for causality )

modules/ROOT/pages/http/http-listener-ref.adoc
dukesphere: "since" -> "because"

modules/ROOT/pages/http/http-listener-ref.adoc
dukesphere: suggest replacement with something like this:
"If you are an advanced user, you can configure persistent connections and set timeouts for them."  
(typical practice is to try to refer to the audience as "you" whenever possible)

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
-* Path = `account/(...)/main-contact`
+When you deploy locally for testing purposes, setting the host to *localhost* is better since it means only requests generated inside the server (your local machine) will be received: the app is not vulnerable to external threats.
+===
 
-=== Wildcard in Path
+Advanced users can also define whether the connections will be persistent and if not, the timeout they will have.
+
+== HTTP Listener
+
+The HTTP listener is configured as a source within flows and must reference the configuration it will use and the path where it will listen for requests.
```
dukesphere: "a source" -> "an Event source" 

"it will use" -> "to use" 

"path where it will listen for" -> "path where it listens for"

modules/ROOT/pages/http/http-listener-ref.adoc
dukesphere: "is a source that allows setting up" -> 
"is an Event source that allows " 
or maybe
"is a Mule Event source that enables you to set up ... and trigger ..."

modules/ROOT/pages/http/http-listener-ref.adoc
dukesphere: ", and on which path requests will be accepted on, therefore allowing routing" ->
"and on which paths to accept requests, thereby allowing the routing of .."

modules/ROOT/pages/http/http-listener-ref.adoc
dukesphere: "source" -> "Event source"

"query params" -> "query parameters"

modules/ROOT/pages/http/http-listener-ref.adoc
dukesphere: "When the flow .... returned, for example" ->  
"You can configure the HTTP listener to customize the HTTP response that occurs when the flow finishes its execution. The customization might return one status code upon successful execution, another upon a failure." 

(I can explain that suggested mod in more detail elsewhere.)

"codes may be" -> "codes can be"

modules/ROOT/pages/http/http-listener-ref.adoc
dukesphere: "a configuration must be declared with a" -> 
"you need to declare a configuration with a" 
(active voice, present tense)

"This establishes" -> "The declaration establishes"  
(we always say what "this" refers to)

modules/ROOT/pages/http/http-listener-ref.adoc
dukesphere: "a base path can be configured that will apply to all ..." -> 
"you can configure a base path that applies to all..." 
(active voice, present tense as much as possible/reasonable)

modules/ROOT/pages/http/http-listener-ref.adoc
dukesphere: "set up as well as" -> "set up, as well as"  
(fyi: I'm leaving future tense "will be set up" above in because it seems that you want to make it clear that a server that wasn't there before actually gets set up as part of this process)

modules/ROOT/pages/http/http-listener-ref.adoc
dukesphere: "... HTTPS, a TLS Context should be defined... details" -> 
"... HTTPS, define a TLS context in the connection, and provide a key store for the server. You can also provide a trust store if you need two-way authentication. See xref:4.1@mule-runtime/tls-configuration.adoc[TLS Configuration] for details."

modules/ROOT/pages/http/http-listener-ref.adoc
dukesphere: "is matched to" -> "matches" (present tense when possible)

"payload while all other metadata" -> "payload, while all metadata"

modules/ROOT/pages/http/http-listener-ref.adoc
dukesphere: "it's result" -> "its result"

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
-In the Path setting, you can use the wildcard asterisk (*) as follows:
+When an HTTP request is matched to a listener, its request data is used to trigger the flow. The body is set as the payload while all other metadata from the request line to the headers are shared as attributes.
+When the triggered flow is done, the HTTP listener uses it's result to generate a proper HTTP response.
 
-* To listen for incoming requests to a path within the specified base path. Use a wildcard at the end of a path.
-* To listen for all requests for a resource in a URI. Use a wildcard in the middle of a path.
+=== About Paths
 
-You can specify a partial path that ends with an asterisk, for example: `/somepath/*`. Any request succeeds that has a URL consisting of the prefix to the asterisk (`/somepath`).
+The path of an HTTP listener can be static, which requires exact matches, or feature placeholders. Placeholders can be wildcards (`*`), which will match against anything it's compared to, or parameters (`{param}`), which will not only match against anything but also capture those values on an URI parameters map.
```
dukesphere: "matches, or" -> "matches or"

"which will match" -> "which match" (present tense)

"it's" -> "they are" 

"which will not only" -> "which  not only" 

"an URI" -> "a URI"

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
 
-Put a wildcard at the end of a path to prevent unexpected results when specific paths configured in other flows override the wildcard path. For example, the Path settings for the HTTP Listener in flows are set as follows:
+Take the following example paths for three listeners using a configuration that establishes `api/v1` as base path:
```
dukesphere: "as base" -> "as the base"

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
 
-* Path configured in Flow A: `/regions/*/customers`
-* Path configured in Flow B: `/regions/us/prospects`
+* `account/mulesoft/main-contact`: will only match the exact path request `http://awesome-company.com/api/v1/account/mulesoft/main-contact`
```
dukesphere: "will only match" -> "only matches"   (present when possible)

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
+* `account/{accountId}/main-contact`: will match all path requests structured similarly such as `http://awesome-company.com/api/v1/account/salesforce/main-contact` and save `salesforce` as the value of `accountId`.
```
dukesphere: "will match" -> "matches"

"similarly such as ... and" -> "similarly, such as ... , and"

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
+* `account/{accountId}/*`: will match all path requests different from `main-contact` such as `http://awesome-company.com/api/v1/account/mulesoft/users` and save `mulesoft` as the value of `accountId`.
```
dukesphere: "will match" -> "matches" 

"such as ... and" -> ", such as ... , and"

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
 
-A request to `/regions/us/customers` fails and is routed to flow B.
+[TIP]
+Wildcards at the end of a path can help capture requests to unhandled resources to provide better error messages.
 
-Put a wildcard in the middle of a path when you want a request to succeed that has a URL consisting of the prefix and suffix to the asterisk. For example, Path = `account/\*/main-contact`. The HTTP Listener listens for every HTTP request having a URI that starts with `/account/` and ends with `/main-contact`; the segment in the middle can vary. Requests succeed from the following URLs:
+When multiple listeners are defined, requests are always routed to the most specific one. Hence in the previous examples a request with `accountId` `mulesoft` and suffix `main-contact` will be handled by the first listener,
```
dukesphere: "Hence in the previous examples a" -> "Hence, in the previous examples, a"  

"`accountId` `mulesoft`" -> "`accountId`, `mulesoft`"

"will be handled" -> "are handled"

modules/ROOT/pages/http/http-listener-ref.adoc
dukesphere: "Request" -> "Requests"

"methods but these" -> "methods, but the methods"

"will be routed" -> "are routed"

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
+and any different `accountId` value by the second listener instead of the third.
 
-* `+http://localhost:8081/account/mulesoft/main-contact+`
-* `+http://localhost:8081/account/foo/main-contact+`
+=== About Methods
 
-=== Using a Placeholder in a URI
+Request can also be routed based on the HTTP method received. By default, a listener will handle all methods but these can be restricted. In the following example, `GET`, `POST` and `PUT` requests will be routed to different flows.
+This is useful to restrict the visibility of data: we may allow certain users to see our data but only a few to modify it.
```
dukesphere: "to restrict" -> "for restricting"

"we may allow .... our data " -> "you might allow ... your data"

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
+</flow>
+
+<flow name="main-contact-general">
+  <http:listener path="account/{accountId}/main-contact" config-ref="HTTP_Listener_config"/>
+  <!-- not supported message -->
+</flow>
+----
 
-For example, a request to `+http://localhost:8081/account/mulesoft/main-contact+` maps the URI parameter to the Mule Message that holds the key `accountId` = `mulesoft`:
+When multiple listeners are defined, requests are routed to the first match in terms of their method, so default listeners should always be defined last.
```
dukesphere: "first match in terms of their method" -> "first matching method"  maybe

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
 
-== Connector Mapping of Received Data
+=== From HTTP Request to Mule Message
 
-The data received by the connector during the Listener operation is always mapped as the payload of the propagated Mule Message.
+The metadata from the HTTP request line, including the method, request path, query and URI params as well as the headers are all available as attributes within a triggered flow.
```
dukesphere: "params as well as" -> "parameters, as well as"

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
+On the other hand, the body and the `Content-Type` header are used to set up the payload and its type. This allows other components to inspect the payload type: DataWeave for example, won't need any input information to work with an HTTP payload.
```
dukesphere: "type. This" -> "type, which"

"DataWeave for example," -> "DataWeave, for example,"

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
+Additionally, if an `X-Correlation-ID` or `MULE_CORRELATION_ID` (for compatibility) header is present, it will be used to set the message's correlation ID for traceability.
```
dukesphere: "if any ... is present, it will be used to" -> "any ... is used to"

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
+  "surname": "Doe",
+  "role": "Senior Vice President",
+  "organization": "Marketing",
+  "phone": 701222369,
+  "email": "john.doe@salesforce.com"
+}
+----
 
-The body of a request is the following HTML form:
+The message's correlation ID will be `9cf32672-4f0b-4e8b-b988-40c13aae85b4` and its payload will be a JSON of:
```
dukesphere: "JSON of:" -> "JSON:"

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
+  "name": "John",
+  "surname": "Doe",
+  "role": "Senior Vice President",
+  "organization": "Marketing",
+  "phone": 701222369,
+  "email": "john.doe@salesforce.com"
+}
+----
+
+An expression like `#[payload.name ++ ' ' ++ payload.surname]` will return `John Doe` since DataWeave will correctly interpret the JSON data.
```
dukesphere: "since" -> "because"

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
+
+The attributes of the message will include:
+
+* method: `POST`
+* listenerPath: `api/v1/account/{accountId}/main-contact`
+* requestPath: `api/v1/account/salesforce/main-contact`
+* relativePath: `account/salesforce/main-contact`
+* queryParams: a multi map with entries `overwrite => true`, `notify => jane.doe` and `notify => admin`
```
dukesphere: "multi map" -> "multi-map" or "multimap" ?

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
+* uriParams: a map with entry `accountId => salesforce`
+* headers: a multi map with entries `host => localhost:8081`, `content-type => application/json`, `content-length => 166` and `x-correlation-id => 9cf32672-4f0b-4e8b-b988-40c13aae85b4`
```
dukesphere: "multi map" -> "multi-map" or "multimap" ?

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
+
+[NOTE]
+Multi maps are similar to maps except they allow several values for a given key. They will return the first value when using a single value selector (`.`) but allow all values to be retrieved using the multiple value selector (`.*`).
```
dukesphere: "Multi maps" -> "Multi-maps" or "Multimaps" ?

"single value" -> "single-value"

"multiple value" -> "multi-value" ?

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
+
+An expression like `#['Received a ' ++ attributes.method ++ ' request for account ' ++ attributes.uriParams.accountId ++ '. The following users will be notified: ' ++ (attributes.queryParams.*notify joinBy ', ')]` will return `Received a POST request for account salesforce. The following users will be notified: admin, jane.doe`.
+
+See the xref:http/http-documentation.adoc[HTTP Documentation Reference] for the complete and detailed listing of HTTP request attributes.
```
dukesphere: "See the" -> "See" (without "the")

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
+
+==== Multipart Form Example
+
+The following example shows the manipulation of an HTTP Listener source message when an HTML form is received, such as:
```
dukesphere: "Listener" -> "listener" (I think it's lower case elsewhere)

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
 blue
- --AaB03x
+--489691234097965980223899--
 ----
 
-== Using DataWeave to Access the Form Data
-
-You access the form values using expressions that resemble these:
+Each item can be accessed through the `parts` object, either by name or item number. For example, the second part can be accessed with `payload.parts.logo` or `payload.parts[1]`. The latter is useful when a name is not provided.
+Within each part, you can access its content and headers. For example, `payload.parts.color.content` will return `blue` while `payload.parts.logo.headers.'Content-Type'` will return `application/png`.
```
dukesphere: "`blue` while" -> "`blue`, while"

recommend using present tense where it makes sense in cases above and below

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
+A very common scenario is obtaining the filename of a part, for this reason, the `Content-Disposition` header is parsed to allow an expression like `payload.parts.logo.headers.'Content-Disposition'.filename` which in this case will return `MuleSoft_logo.png`.
```
dukesphere: "part, for" -> "part. For"  (those are 2 independent clauses)

"'Content-Disposition'.filename` which" -> "'Content-Disposition'.filename`, which "

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
 
-* Access the content of the `name` field
-+
-`#[payload.name]`
-+
-* Access the headers
-+
-`#[attributes.headers['content-type']`
+See the xref:4.1@mule-runtime/dataweave-formats.adoc[Formats Supported by DataWeave] to learn more about reading and writing multipart content.
```
dukesphere: "See the" -> "See" (without "the")

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
 
-For more information, see "About HTTP Responses to Requests."
+=== About HTTP Responses
 
-== HTTP Custom Properties Request Line
+After the triggered flow is done processing, its result is returned so that the listener can provide a response. If the flow executed successfully, a regular response will be returned but if the flow terminated with a failure then an error response will be returned.
```
dukesphere: "is done processing" -> "finishes processing" 

"returned but" -> "returned, but" 

"with a failure then an error" -> "with a failure, an error"

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
+By default, a regular response will return a 200 status code and the message payload as body, while an error response will return a 500 status code with the flow error's description as body. These responses can be customized by providing:
```
dukesphere: "as body" -> "as the body" maybe, and both instances

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
 
-Besides headers and a body, an HTTP request is also composed of a request line. The HTTP request line is the content on the first line of the HTTP Request that includes the URI from the address bar of the client request. For example:
+* the status code
+* the reason phrase
+* the headers multi map
```
dukesphere: "multi map" -> "multi-map" or "multimap" ?

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
+* the body
 
-`POST /mydomain/login/?user=aaron&age=32 HTTP/1.1`  +
+DataWeave can be used to generate each one and variables can be used to propagate data from the flow.
```
dukesphere: "each one and" -> "each one of these responses, and" ?

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
 
-This content is transformed into the following attributes in the Mule Message:
+In the following example, an endpoint is defined to store logos for an account.
 
-Property names are defined as follows:
+When the storage is successful, a `201 Created` is returned with a body such as `Corporate Logo has been stored as a mulesoft logo.`.
```
dukesphere: "body such as" -> "body, such as"

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
-The path: `/mydomain/login/`
-+
-* `#[attributes.queryParams.user]` and `#[attributes.queryParams.age]`
-+
-The query string values from the URI, `aaron` and `32`.
-+
-* `#[attributes.queryString]`
-+
-The query string from the URI, without parsing, `user=aaron&age=32`.
+When the storage fails, the status code is defined through the `errorCode` variable if available or `500` by default. Additionally, a custom header is added (`X-Time`) and a body such as `Corporate Logo could not be stored.` is returned.
```
dukesphere: "if available" -> "(if available)" maybe or ", if available," prob former though

"a custom header is added (`X-Time`) and a body such as  ... is returned." -> 
"a custom header (`X-Time`) is added, and a body, such as ..., is returned."

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
+Should there be a `CONNECTIVITY` error storing the logo, for example, the returned status code would be `504` while any other errors would result in a `500`.
```
dukesphere: "Should there be a" -> "If there is a"  ( the latter is more common )

"`504` while" -> "`504`, while"

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
 
-When the HTTP Listener has URI parameter placeholders in its path, a Map is created with the holder name as key and the value as extracted from the request path. For example, `domain=mydomain`. The Listener path must be defined as `{domain}/login`
+It's important to note that while a reason phrase is not defined for error responses,
+ the connector will attempt to define it based on the status code. Thus, a `Gateway Timeout` or `Internal Server Error` will be returned in the scenarios explained before.
```
dukesphere: "it" -> "one"

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
+----
 
-The response message sent by the HTTP Listener can include HTTP headers. In Headers, enter an expression to include information in the response, for example: `#[{'Connection' : 'close' }]`. This will add a `Connection` header with `close` value to the HTTP response.
+==== HTTP Streaming Mode
 
-=== HTTP Response Status Code and Reason Phrase
+When handling response bodies, the HTTP connector will consider the kind of data to send and use `chunked` encoding when the size is not clear (think streams with no size information). You can change this behavior using the `responseStreamingMode` options:
 
-You can customize the connector response to requests dynamically. For example, you can set different error status code numbers depending on the error raised by having an error handler that sets the value of a variable used to define the HTTP response status code.
+* AUTO (default): if a size is defined for the body it will use `Content-Length` encoding, otherwise it will use `Transfer-Encoding: chunked`.
```
dukesphere: "body it" -> "body, the connector" (safer/clearer to say what "it" refers to)

"encoding, otherwise" -> "encoding. Otherwise, "

modules/ROOT/pages/http/http-listener-ref.adoc
dukesphere: Just FYI: "can be done" is passive voice. Try to use active voice in tech writing whenever possible. It's a standard practice in the industry. Active voice tells the reader who or what is doing the action. 

See my suggestion below.

modules/ROOT/pages/http/http-listener-ref.adoc
marianogonzalez: source -> not sure all readers will get this one as the `message source` is a term we're kinda moving away from in our public docs. Consider replace it by `trigger` or at least `event source`

modules/ROOT/pages/http/http-listener-ref.adoc
marianogonzalez: what about the other verbs? the term `both` seams to imply that no other verb is supported.

modules/ROOT/pages/http/http-listener-ref.adoc
marianogonzalez: `Once a request is accepted by the source` -> `Once a request is accepted by the listener`

marianogonzalez: `source` -> `listener`

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
+To use an HTTP listener, a configuration must be declared with a corresponding connection. This establishes the HTTP server that will listen to requests.
 
-== Setting the Host
+Additionally, a base path can be configured that will apply to all listeners using the configuration.
 
-To deploy a project to CloudHub, setting Host = *0.0.0.0* is required.
+[source,xml,linenums]
+----
+<http:listener-config name="HTTP_Listener_config" basePath="api/v1">
+  <http:listener-connection host="0.0.0.0" port="8081" />
```
marianogonzalez, line 24: since it's a common practice, should we add a comment around the difference between `0.0.0.0` and `localhost`. I know that this doc is about the http connector and not basic networking, but still......

modules/ROOT/pages/http/http-listener-ref.adoc
marianogonzalez: I don't understand the point of this. If I'm deploying locally for testing, which threats are you talking about? Isn't this a concern of when you actually deploy to production or to an external staging environment at most?

modules/ROOT/pages/http/http-listener-ref.adoc
marianogonzalez: not really clear. 
* `feature placeholders` -> I get what you mean but not sure that's a clear/common way of referring to that. Maybe use `path params`?
* `which will match against anything it's compared to` -> not clear

modules/ROOT/pages/http/http-listener-ref.adoc
marianogonzalez: Don't be support the OPTIONS method too? I was also under the impression we even support custom methods.

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
-=== Using a Placeholder in a URI
+Request can also be routed based on the HTTP method received. By default, a listener will handle all methods but these can be restricted. In the following example, `GET`, `POST` and `PUT` requests will be routed to different flows.
+This is useful to restrict the visibility of data: we may allow certain users to see our data but only a few to modify it.
 
-You can define a path that uses a placeholder in curly braces: `{accountId}` for example, `account/{accountId}/main-contact`. The connector maps the accountId to a URI parameter attribute in the flow.
+[source,xml,linenums]
+----
+<flow name="main-contact-write">
+  <http:listener path="account/{accountId}/main-contact" allowedMethods="POST, PUT" config-ref="HTTP_Listener_config"/>
+  <!-- validate user permissions -->
```
marianogonzalez, line 93: would be good to actually give an example implementation on how to do these two things

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
+</flow>
+
+<flow name="main-contact-read">
+  <http:listener path="account/{accountId}/main-contact" allowedMethods="GET" config-ref="HTTP_Listener_config"/>
+  <!-- fetch main contact for accountId -->
+</flow>
+
+<flow name="main-contact-general">
+  <http:listener path="account/{accountId}/main-contact" config-ref="HTTP_Listener_config"/>
+  <!-- not supported message -->
```
marianogonzalez: not clear. If this comment to say "this will never work" or is a comment to say "here you send an error"? If:

1) then this is not clear and I would put this as a separate example saying "notice this won't work because ABC, and the following error will be raised
2) If this is the case, then show how to raise an error, sending a custom error body and status code.

modules/ROOT/pages/http/http-listener-ref.adoc
marianogonzalez: I wouldn't refer to attributes as metadata. I know we colloquially do internally, but for users it might get confused with DataSense. I would look for a replacement terms as `request data` vs `request body`, or even just plainly `attributes`. This comment actually applies to the entire thing. Sorry :)

modules/ROOT/pages/http/http-listener-ref.adoc
marianogonzalez: `and its type` -> and its mimeType. The actual type of data is up to us.

modules/ROOT/pages/http/http-listener-ref.adoc
marianogonzalez: shoudn't this link to some other article explaining what the correlation Id is?

marianogonzalez: `for compatibility` -> for interoperability with Mule 3

modules/ROOT/pages/http/http-listener-ref.adoc
marianogonzalez: this should link to some reference of what a Multimap is and how to use it. The term will not automatically resonate with most users.

Also, that link should contain examples of how to access it. E.g: 

1) `vars.multimap.bleh` returns the first element
2)`vars.multimap.bleh[1]` returns the second element 
3) `vars.multimap.*bleh` returns all
4) `sizeOf(vars.multimap)` well.. you get the picture

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
 
 blue
- --AaB03x
+--489691234097965980223899--
 ----
 
-== Using DataWeave to Access the Form Data
-
-You access the form values using expressions that resemble these:
+Each item can be accessed through the `parts` object, either by name or item number. For example, the second part can be accessed with `payload.parts.logo` or `payload.parts[1]`. The latter is useful when a name is not provided.
```
marianogonzalez, line 217: maybe add examples of:

1) I have N parts, they all specify a json mimeType
2) I have N parts, they all hold json payloads but they don't specify it in the request
3) I have N parts, they are all of different mimeTypes, some specified, some not.

modules/ROOT/pages/http/http-listener-ref.adoc
marianogonzalez: `is done processing` is ambiguos, as the control is also returned to the listener is the flow fails.

modules/ROOT/pages/http/http-listener-ref.adoc
marianogonzalez: a common ask is how to set the response code based on some condition. Would be good to add a simple example of that. (e.g: `statusCode="#[vars.recordInserted ? '201' : '200']"` , or however a ternary is written in DW)

modules/ROOT/pages/http/http-listener-ref.adoc
```diff
@@ -1,182 +1,296 @@
...
-After request processing, the message is returned to the HTTP connector to provide a response. You can customize the response body, attachment, headers, and status.
-
-=== HTTP Response Body
-
-The HTTP Response body is generated from payload of the Mule Message after executing the flow unless explicitly configured. You can use a DataWeave script or expression to customise it.
-
-=== HTTP Response Headers
+[source,xml,linenums]
+----
+<flow name="store-logo">
```
marianogonzalez, line 252: for this kind of large flows, you can use numeric callouts to add line by line explanations below. Check the SDK docs, it's full of examples of this and we found it really helps readability.

marianogonzalez (changes requested):

modules/ROOT/pages/http/http-listener-ref.adoc
afelisatti: "You can configure the HTTP listener to customize the HTTP response that occurs when the flow finishes its execution. The customization might return one status code upon successful execution, another upon a failure." -> I don't think this works. The important thing to note here is that once the flow finishes, the listener responds. It's not the customization, that is secondary.

modules/ROOT/pages/http/http-listener-ref.adoc
afelisatti: There's a tip around this on the connection section where this config is repeated.

modules/ROOT/pages/http/http-listener-ref.adoc
afelisatti: exactly, I want to highlight the fact that this configuration creates an actual server.

modules/ROOT/pages/http/http-listener-ref.adoc
afelisatti: It refers to the fact that using `0.0.0.0` locally while testing will allow external connections into the machine.

modules/ROOT/pages/http/http-listener-ref.adoc
afelisatti: `More advanced configurations can also define whether the connections will be persistent and if not, the timeout they will have.`

modules/ROOT/pages/http/http-listener-ref.adoc
afelisatti: `path params` is confusing since the term `URI params` is already established and similar. I don't see how to make this clearer other than with examples as the ones below.

afelisatti: `"matches, or" -> "matches or"` without the comma it would sound like a static path requires an exact match or feature placeholders.

afelisatti: Perhaps just saying the path can be static or parameterized?

modules/ROOT/pages/http/http-listener-ref.adoc
afelisatti: `"accountId, mulesoft"` I don't think this makes sense, I'm trying to say the `accountId` parameter will be `mulesoft`.

modules/ROOT/pages/http/http-listener-ref.adoc
afelisatti: We do. What part is unclear? I chose a usual example but it shouldn't imply that's all we support.

modules/ROOT/pages/http/http-listener-ref.adoc
afelisatti: I don't agree, even if I add an excellent example no one will remember to update the related components used here. The examples will be destined to become obsolete and then we'll have users asking questions about the example they copied and are not working. 
If this was a mule cookbook like the one DW has it would make sense, but this is HTTP exclusive.

modules/ROOT/pages/http/http-listener-ref.adoc
afelisatti: I'll add a set payload since it's standard.

modules/ROOT/pages/http/http-listener-ref.adoc
afelisatti: there's no `any` in the sentence 🤔

afelisatti: I couldn't find one.

modules/ROOT/pages/http/http-listener-ref.adoc
afelisatti: I added a note on this but will create a jira to add a full doc on it.

modules/ROOT/pages/http/http-listener-ref.adoc
afelisatti: I don't understand what the difference is if they specify a mime type or not, all are the same or not, etc. The expressions to access the data are always the same. Or are you saying to explain what a lack of content-type means?

modules/ROOT/pages/http/http-listener-ref.adoc
dukesphere: Event source is perfect! :) You see "Source" in Studio, which is too overloaded with other meanings to be useful by itself in the docs and will be impossible to translate accurately. "Trigger" is what you see in Flow Designer. It would be ideal if the two UIs could use the same terminology.

modules/ROOT/pages/http/http-listener-ref.adoc
dukesphere: ok, the separate set of backticks (around each term) led me to think they were supposed to be separate things.

dukesphere (approved):

marianogonzalez (approved):

PR #96, opened 2019-01-25
Fixed broken links in table

george-hoffer:
Also added blocking attribute to Reconnect and Reconnect Forever per https://www.mulesoft.org/jira/browse/DOCS-3789 (needs to be added to all Mule 4 connector reference docs)


wwarner-salesforce (approved):
Looks good!

PR #97, opened 2019-01-25
DOCS-3304 update on Spring examples

george-hoffer:
https://www.mulesoft.org/jira/browse/DOCS-3304


wwarner-salesforce (approved):
Looks good!

PR #98, opened 2019-01-25
DOCS-2942 fix to remove toc entries at top

george-hoffer:
Plus updated Reconnect and Reconnect Forever per DOCS-3789


wwarner-salesforce (approved):
Looks good!

PR #100, opened 2019-01-29
Fix links httpref duke

dukesphere:
fixing some links, I tested the fixes locally


valkyrie69 (approved):
I assume you verified all links.  :-)

PR #101, opened 2019-01-30
Updated popular connector list

george-hoffer:
Also added link to Microsoft Windows Gateway Services (line 74) and changed Support link to the KB link in line 118.


wwarner-salesforce (approved):
Looks good!

PR #108, opened 2019-02-12
DOCS-4121 build fixes

george-hoffer:
Fixing files so they won't appear as local build warnings.


wwarner-salesforce (approved):
Looks great! Lots of fixes in this one.

PR #111, opened 2019-02-12
DOCS-3844 fix

george-hoffer:
https://www.mulesoft.org/jira/browse/DOCS-3844
The previously _partial reference to https://raw.githubusercontent.com/mulesoft/docs-connectors/latest/modules/ROOT/pages/_partials/common-streaming.adoc was the error that DOCS-3844 is about -- Using the "List operation" in the _partial file was completely incorrect in the context of thae Read operation. Solution was to copy the partial text to this file and change "List" to "Read".


wwarner-salesforce (approved):
Looks good!

PR #115, opened 2019-02-14
Docs 4163 - Add "Cached OAuth Username Password" table 

george-hoffer:
https://www.mulesoft.org/jira/browse/DOCS-4163 

Added new "Cached OAuth Username Password" table obtained from a diff of the connector references for 9.5.0 and 9.5.1. 

Each reference was created by checking out the version branches from https://github.com/mulesoft/mule-salesforce-connector, and running "mvn clean package -DskipTests" in the root of each to generate a version of the connector reference.


wwarner-salesforce (approved):
Looks good!

PR #116, opened 2019-02-15
Docs 4163 Salesforce Reference

george-hoffer:
Reformatted table because required fields couldn't be seen in the rightmost column.


wwarner-salesforce (approved):
Looks good!

PR #119, opened 2019-02-20
Light edits for aggregators documentation

lucianoRM:


modules/ROOT/pages/aggregator/aggregator-examples.adoc
```diff
@@ -65,15 +65,52 @@ But what if you do not want to wait forever to fill every batch. In that case, y
...
 <aggregators:aggregator-listener aggregatorName="personsNameAggregator" includeTimedOutGroups="true"/>
 <flow-ref name="whatToDoWithAllTheseNames"/>
 ----
 
-Keep in mind that for this case, the data types that come from the `aggregation-complete` route and `aggregator-listener` are the same. So, for the execution of the flow `whatToDoWithAllTheseNames`, it does not matter where the data comes from.
+For this case, the data types that come from the `aggregation-complete` route and `aggregator-listener` are the same. So, for the execution of the flow `whatToDoWithAllTheseNames`, it does not matter where the data comes from.
+However, with this configuration, in the case the aggregation is complete, the flow `whatToDoWithAllTheseNames` will trigger twice, once from the `aggregation-complete` route `flow-ref` and again from the `flow-ref` in the flow with the `aggregator-listener`.
+
+To avoid multiple executions of the same logic, the `flow-ref` after the `aggregator-listener` should be wrapped in a `choice` component that uses the `isAggregationComplete` attribute to check for an incomplete aggregation. That way, the `flow-ref` is only called when a timeout triggered that flow.
```
dukesphere, line 79: I made this change: "To avoid multiple executions of the same logic, the `flow-ref` after the `aggregator-listener` should be wrapped in a `choice` component that uses the `isAggregationComplete` attribute to check for an incomplete aggregation. That way, the `flow-ref` is only called when a timeout triggered that flow." Are you ok with it?

dukesphere (approved):
Approving but see my question about this change:

modules/ROOT/pages/aggregator/aggregator-examples.adoc
lucianoRM: Yes! i like it.

lucianoRM: Please merge.

PR #120, opened 2019-02-21
DOCS-4194 added ref to forum article

george-hoffer:
https://www.mulesoft.org/jira/browse/DOCS-4194


wwarner-salesforce (approved):
Looks good!

PR #126, opened 2019-03-11
DOCS-4345 - updated reference to 2.1.2

george-hoffer:
https://www.mulesoft.org/jira/browse/DOCS-4345 -- This reference has been built locally and fixes applied.


wwarner-salesforce (approved):
Looks good!

PR #131, opened 2019-03-16
DOCS-4307 Change filename-pattern to filenamePattern

george-hoffer:
Super quick fix changing "filename-pattern" to "filenamePattern" per https://www.mulesoft.org/jira/browse/DOCS-4307


wwarner-salesforce (approved):
Looks great!

PR #132, opened 2019-03-16
IBM MQ Reference: Block syntax fix

rhoegg:
for IBM MQ Attributes, JMS Message Properties, and JMSX properties


valkyrie69:
Put changes in https://github.com/mulesoft/docs-connectors/pull/507

PR #146, opened 2019-04-05
correct field key name

wwarner-salesforce:
In #docs-feedback today support engineer Randall Tom said the correct field key name is runServerSuiteScriptAndTriggerWorkflows.


wwarner-salesforce:
George demonstrated that the XML value is "runServerSuiteScriptAndTriggerWorkflows", but the NetSuite Config user interface displays "Run server SuiteScript and trigger workflows", and for all field keys the reference document shows UI text and does not show XML values.

george-hoffer (approved):

PR #161, opened 2019-04-13
Updated names of Amazon Connectors

amitfernandes:
Updated names of Amazon Connectors to include the word "Amazon" in the name. Example - Amazon S3


amitfernandes:
Updated names of Amazon Connectors to include the the word "Amazon" in the name. Example - Amazon S3

wwarner-salesforce (approved):
Thank you!

PR #166, opened 2019-04-17
DOCS-4721 ms sb ref update

george-hoffer:
https://www.mulesoft.org/jira/browse/DOCS-4721


wwarner-salesforce (approved):
Looks great! Thank you very much for helping with this.

PR #167, opened 2019-04-19
Update mongodb-connector-reference.adoc

george-hoffer:
Tested in local build and good to go.

### DIFF
--- a/modules/ROOT/pages/mongodb/mongodb-connector-reference.adoc
+++ b/modules/ROOT/pages/mongodb/mongodb-connector-reference.adoc
@@ -1434,7 +1434,7 @@ Inserts a list of documents in a collection, setting its id if necessary.
 | Configuration | String | The name of the configuration to use. | | x
 | Collection Name a| String |  +++The name of the collection where to insert the given document.+++ |  | x
 | Documents a| Binary |  +++A java.util.List of String.+++ a|  `#[payload]` | 
-| Write Ordered a| Boolean a|  +++Indicates if the list of write operations is ordered or unordered. By default, if an error occurs during the processing of one of the write operations, MongoDB returns without processing any remaining write operations in the list. Returns new `BsonDocumentWrapper(this, codecRegistry.get(Document.class));`+++ |  +++true+++ | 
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/mongodb/mongodb-connector-reference.adoc
```diff
@@ -14,2281 +18,2417 @@ MongoDB is an open source, high-performance, schema-free, document-oriented data
...
 
 
 [[createFileFromPayload]]
-== Create File From Payload
+=== Create File From Payload
 `<mongo:create-file-from-payload>`
 
 +++
-Creates a new MuleGridFSFile in the database, saving the given content, filename, contentType, and extraData, and answers it.
+Creates a new MuleGridFSFile in the database, saving the given content, filename, contentType, and extraData, and answers. it.
```
wwarner-salesforce: I don't understand this. Answers what? Maybe it should be "answers in it." meaning that some sort of answers are part of the created file? Or maybe it answers the call to create the file by sending back a copy of the file, or a link to it?

george-hoffer: Good catch. I removed "it." I think the idea is that the content that's saved includes answers to some inquiry.

wwarner-salesforce: OK, sounds good.

modules/ROOT/pages/mongodb/mongodb-connector-reference.adoc
```diff
@@ -14,2281 +18,2415 @@ MongoDB is an open source, high-performance, schema-free, document-oriented data
...
+* MONGO:MONGO_GRID_FS 
+* MONGO:SOCKET_OPEN 
+* MONGO:CONNECTIVITY 
+* MONGO:EXECUTION_TIMEOUT 
+* MONGO:SECURITY 
+* MONGO:MAPPING_EXCEPTION 
+* MONGO:NODE_IS_RECOVERING 
+* MONGO:INTERNAL_EXCEPTION 
+* MONGO:SERVER_ERROR 
+* MONGO:DUPLI
```
wwarner-salesforce: I'm wondering if the formatting here is correct. Maybe it should be something like this?
Example: `<$set,{"key":123}>`

george-hoffer: Changed line 2156 to:
For example: `<$set,{"key":123}>`

wwarner-salesforce (approved):

PR #171, opened 2019-04-23
DOCS-4769 - Update Redis Reference

george-hoffer:
https://www.mulesoft.org/jira/browse/DOCS-4769

### DIFF
--- a/modules/ROOT/pages/redis/redis-connector-reference.adoc
+++ b/modules/ROOT/pages/redis/redis-connector-reference.adoc
@@ -3,1223 +3,1102 @@ ifndef::env-site,env-github[]
 include::_attributes.adoc[]
 endif::[]
 
+Support Category: Select
+
+Redis Connector v5.1
+
 == Configurations
 ---
 [[redis]]
 === Redis
 
 
-=== Parameters
-
-[cols=".^20%,.^20%,.^35%,.^20%,^.^5%", options="header"]
+==== Parameters
+[%header%autowidth.spread]
 |===
 | Name | Type | Description | Default Value | Required
 |Name | String | The name for this configuration. Connectors reference the configuration with this name. | | x
-| Connection a| * <<redis_clustered, Clustered>>
-* <<redis_nonclustered, NonClustered>>
- | The connection types that can be provided to this configuration. | | x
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/redis/redis-connector-reference.adoc
```diff
@@ -3,1264 +3,1145 @@ ifndef::env-site,env-github[]
...
 | Configuration | String | The name of the configuration to use. | | x
 | Key a| String |  Key to use for LPUSH/RPUSH/LPUSHX/RPUSH |  | x
-| Value a| String |  The value to push. |  #[payload] |
-| Push Left a| Boolean |  The side where to push the payload, either LEFT or RIGHT |  false |
-| If Exists a| Boolean |  If true execute LPUSHX/RPUSH otherwise LPUSH/RPUSH |  false |
-| Target Variable a| String |  The name of a variable that stores the output from the operation. |  |
-| Target Value a| String |  An expression to evaluate against the operation's output and the outcome of that expression is stored in the target variable |  #[payload] |
+| Value a| String |  The value to push. |  `#[payload]` | 
+| Push Left a| Boolean |  Side that receives the pushed payload. If the parameter is true, push the payload on the left side. If the parameter is false, push the payload on the right side. |  false | 
+| If Exists a| Boolean |  If the parameter is true, execute LPUSHX or RPUSH. If the parameter is false, execute LPUSH or RPUSH. |  false |
```
wwarner-salesforce: I think the note on this line is worth discussing with a dev. First, I wonder if maybe it should be "LPUSHX/RPUSHX" for true and "LPUSH/RPUSH" for false, and the dev made a tiny and extremely dangerous typo. (I do see that line 922 also has "LPUSH/RPUSH/LPUSHX/RPUSH".) Second, I think LPUSH/RPUSH means "LPUSH or RPUSH" but it might mean "LPUSH and then RPUSH". Third, references to functions like LPUSH should probably be replaced with a more English description of the effects. For example, "If the parameter is true and the list exists, push the message payload onto the list. If the parameter is true and the list does not exist, do nothing. If the parameter is false, push the message payload onto the list, creating a new list if one does not exist already." This example may or may not match the actual behavior of the function, which is what I'd like to see explained more fully by a dev.

wwarner-salesforce (approved):
Looks good, although I think line 925 could use some clarification from a dev, either in this PR or another one soon.

modules/ROOT/pages/redis/redis-connector-reference.adoc
george-hoffer: Checking with the devs

george-hoffer: Good catch. Fixed to show LPUSHX or RPUSHX. Thanks!

PR #165, opened 2019-04-17
Twilio 3.0.2

marianborzik:

PR #172, opened 2019-04-23
replace some links on Twilio user guide

wwarner-salesforce:


wwarner-salesforce:
These were removed in https://github.com/mulesoft/docs-connectors/pull/165 but after discussing with George I've decided to restore them.

george-hoffer (approved):

PR #193, opened 2019-05-06
AWS S3 connector 5.3.1 - Added additional info to connector configuration

xidex:
Added additional information to connector configuration using Anypoint Studio to avoid confusion when using Test Connection functionality. (https://www.mulesoft.org/jira/browse/SE-11434)


modules/ROOT/pages/amazon/amazon-s3-connector.adoc
```diff
@@ -71,7 +71,8 @@ src folder of your project.
...
+. Click Test Connection to confirm that the parameters of your global configuration are accurate, and that Mule is able to successfully connect to your instance of Amazon S3. Access to operations on Amazon S3 is controlled through policies for that user. It may not always be possible to validate the credentials before the exact operation for which the user has access is done. This can vary for instance on the bucket name and other parameters. Hence the test connection may fail in case the credentials have a restricted policy such as access to a specific folder within a specific bucket. Validating a connection with Test Connection requires that the user has permission in AWS IAM to the action `s3:ListAllMyBuckets`. If the user does not have this permission then Test Connection will fail.
```
dhirsch1: Question: If a user has permission in AWS IAM to the action `s3:ListAllMyBuckets` will Test Connection always work, even if the user has restricted policies?

xidex: Yes, if a user has `s3:ListAllMyBuckets` permission, the test connection will always work. 
Maybe those two sentences should be reworded. The `s3:ListAllMyBuckets` permission is needed only for connection validation. User can still use the connector and the operations he has access to just fine (if the correct credentials were provided), only the connection validation will not work.

dhirsch1: Here is suggested text. I am not really clear about the last sentence though. Do we need it? I removed the example. It did not really explain why the connection would fail.

Click Test Connection to confirm that the parameters of your global configuration are accurate, and that Mule is able to successfully connect to your instance of Amazon S3. 

Validating a connection with Test Connection requires that you have permission in AWS IAM to the action `s3:ListAllMyBuckets`. If you don’t have this permission, Test Connection fails. However, you can still use the connector and the operations to which you have access. 

Access to operations on Amazon S3 is further controlled through policies. It is not always possible to validate your credentials before the exact operation for which you have access completes. This can vary based the bucket name and other parameters. For example, the test connection can fail if your credentials have a restricted policy.

xidex: I agree with the suggestion and I have updated the text accordingly.

dhirsch1 (approved):

PR #198, opened 2019-05-13
DOCS-4869 update to Salesforce 9.5 ref

george-hoffer:


modules/ROOT/pages/salesforce/salesforce-connector-reference.adoc
```diff
@@ -7,6 +7,9 @@ Support Category: Select
 
 Salesforce Connector version 9.5
 
+This reference document provides Anypoint Studio field information for each Salesforce connector operation. See the
```
valkyrie69, line 10: Should it just be Salesforce Connector 9.5 (no version)

valkyrie69 (approved):

modules/ROOT/pages/salesforce/salesforce-connector-reference.adoc
george-hoffer: There are 2 connector guides, one for v9.5 and for v9.6. Wanted to make sure they went to the correct guide.

PR #200, opened 2019-05-16
Updated 'test connection' note

john-callahan:
Added a note to change the transport configuration on the connector global configuration when trying to test the connection. Default is a bean reference, and the test connection will fail even with correct credentials and authorization to the Get_Workday_Account_Signons_Request operation. This is on Workday Connector v10.0.0, Mule v4.2.0, Studio 7.3.4


mcherem (approved):
Added some doc edits.

PR #201, opened 2019-05-20
remove derby

Mulesoft-Mysti:
derby-connection uri no longer works, so special instructions for this db don't work.


mcherem (approved):
Looks good.

PR #205, opened 2019-05-21
Fix DB Examples

mcherem:
Fix DB Examples

### DIFF
--- a/modules/ROOT/pages/db/db-connector-sp.adoc
+++ b/modules/ROOT/pages/db/db-connector-sp.adoc
@@ -1,11 +1,11 @@
-= Execute Stored Procedures - Mule 4
+= Example: Execute Stored Procedures - Mule 4
 ifndef::env-site,env-github[]
 include::_attributes.adoc[]
 endif::[]
 
-Invoke stored procedures that combine, input, output, and input-output parameters. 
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/db/db-connector-sp.adoc
```diff
@@ -1,11 +1,11 @@
-= Execute Stored Procedures - Mule 4
+= Example: Execute Stored Procedures - Mule 4
 ifndef::env-site,env-github[]
 include::_attributes.adoc[]
 endif::[]
 
-Invoke stored procedures that combine, input, output, and input-output parameters. 
+Create a stored procedures that allows the input, output, and input-output of payload parameters.
```
estebanwasinger, line 6: This doesn't create stored procedures, invokes stored procedures.
Doesn't exist such thing as "payload parameters"

estebanwasinger (changes requested):

estebanwasinger (changes requested):
This PR doesn't fix any of the issues that the documentation have:
As a list of things that I think should be fixed:
1) The layout, one big example is not ok, should be split 
2) Explain what is an input parameter, an output parameter, and in-out parameter
2b) when I should use each of those
3) missing real world examples

mcherem:
This has been superseded by a PR from the Dev team.

PR #212, opened 2019-05-24
Update salesforce-connector-97.adoc

george-hoffer:


DBCSV (approved):

modules/ROOT/pages/salesforce/salesforce-connector-97.adoc
```diff
@@ -206,6 +206,8 @@ NOTE: Before you can subscribe to a channel or topic, it must first be published
 * *Subscribe Channel* - Subscribe to a streaming channel. This trigger provides channel-related notification settings for new events that take place after you subscribe.
 * *Subscribe Topic* - Subscribe to a topic. This trigger provides topic-related notification settings for new events that take place after you subscribe.
 
+IMPORTANT: For the *On New Object* and *On Modified Object* operations, specify a UTC value in the *Since* field. If the server is in a different time zone, you must specify a value for the *Since* field; if the *Since* field is left empty, the connector takes the local timestamp. As servers are in datacenters located generally elsewhere from one's location, use of the *Since* field is recommended.
```
valkyrie69: I don't think the last line is needed at all, but if you feel otherwise, it should be rewritten. Servers in datacenters are often located elsewhere, in different time zone, so specifying a UTC value in the *Since* field is recommended. 
Or similar...however you like. The first sentence already covers it though by making it sound not optional. If it's recommended, then just tell them to do it. 

Or:
You must specify a UTC value in the *Since* field for the On New Object and On Modified Object operations, as severs are often located in different locations from the local time zone, and if the *Since* field is left empty, the connector uses the local timestamp. 

Or something. But really that's all they need to know IMO.

valkyrie69 (approved):
One suggestion that you can take or leave. Not a show stopper.

modules/ROOT/pages/salesforce/salesforce-connector-97.adoc
george-hoffer: Went with:
IMPORTANT: For the *On New Object* and *On Modified Object* operations, specify a UTC value in the *Since* field. If the server is in a different time zone, you must specify a value for the *Since* field; if the *Since* field is left empty, the connector takes the local timestamp. Servers in datacenters are often located elsewhere, in a different time zone, so specifying a UTC value in the *Since* field is recommended.

PR #216, opened 2019-05-29
DOCS-5106 update

george-hoffer:
https://www.mulesoft.org/jira/browse/DOCS-5106

### DIFF
--- a/modules/ROOT/pages/object-store/object-store-connector-reference.adoc
+++ b/modules/ROOT/pages/object-store/object-store-connector-reference.adoc
@@ -48,9 +48,6 @@ A ConnectionProvider that returns the Mule runtime's default ObjectStoreManager.
 | Reconnection a| <<Reconnection>> |  +++When the application is deployed, a connectivity test is performed on all connectors. If set to true, deployment fails if the test doesn't pass after exhausting the associated reconnection strategy.+++ |  |
 |===
 
-
-
-
 == Operations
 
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/object-store/object-store-connector-reference.adoc
```diff
@@ -378,11 +377,9 @@ If not provided, the Mule runtime's default ObjectStoreManager is used. Setting
 ** MINUTES
 ** HOURS
 ** DAYS a| An entry's time to live (`#entryTtl`). | SECONDS |
-| Expiration Interval a| Number | How frequently the expiration thread should run.
-
-If #entryTtl is set, then the expiration thread does not run despite the value set there.
+| Expiration Interval a| Number | How frequently the expiration thread should run. The default value is 1 minute.
 
-If set to a value lower than or equal to zero, then there is no expiration. | 1 |
+The `Expiration Interval` field requires that the `Max Entries` or `Entry Ttl` fields also be set. If neither is set, then putting a value in `Expiration Interval` has no effect and the expiration thread does not run.  | 1 |
```
wwarner-salesforce: Should this section also include this line at the end?
```
If `Expiration Interval` is set to a value lower than or equal to zero, then the object store never expires.
```

wwarner-salesforce (approved):
Left one question. Looks great!

PR #229, opened 2019-06-03
DOCS-5055 fix and edits

george-hoffer:


george-hoffer:
Unclear what the version is. Closing.

PR #230, opened 2019-06-04
Fix db documentation typo

Bardammu:


george-hoffer:
Thank you for your contribution. Please complete the contributor's agreement as described at https://github.com/mulesoft/docs-connectors/blob/latest/README.adoc to register as a contributor so that we can accept your pull request.

Bardammu:
@george-hoffer, what is that? I've been a Mulesoft employed for more than four years, five in two months :)

PR #233, opened 2019-06-06
Docs 4932 ServiceNow DH 

dhirsch1:
I have converted the ServiceNow connector documentation to the new template. Please note the following:

I see that my deletions of the old files did not get into this pull request. Please let me know what I need to do to fix this. Can I just delete them directly in the PR?

I found three examples in the code, although only one was included in the docs before. I added all three but need to check whether the two others are valid.

I was not able to successfully create an xref to the examples in the Get Started topic. I thought it was important to list each example there since examples are what the users have been asking for.

In the Audience section of the Get Started topic, the URLs are breaking up into separate lines. What do we use to indicate a non-breaking space in ASCiiDoc? Would this work in a URL?

I think there are areas where we should be using back ticks and also some capitalization issues. We can go over this after we have converted all of the templates.


dhirsch1:
Works was completed by Maria Cherem.

PR #238, opened 2019-06-12
formatting consistency

nathan-at-mulesoft:


george-hoffer:
Our new style is to list: Support Category: Premium

PR #260, opened 2019-07-08
DOCS-5405 fix

george-hoffer:
Waiting for a tech review from Alexandru Atomei.


george-hoffer:
This PR needs to be redone in light of the new doc formats.

PR #272, opened 2019-07-12
DOCS-5405 Added Get Updated Objects section

george-hoffer:
And put Usage Notes in alphabetic order.

### DIFF
--- a/modules/ROOT/pages/salesforce/salesforce-connector-config-topics.adoc
+++ b/modules/ROOT/pages/salesforce/salesforce-connector-config-topics.adoc
@@ -313,17 +313,43 @@ output application/java
 [[usagenotes]]
 == Usage Notes
 
+=== Currency
+
+* Currency values cannot exceed 18 characters in length.
+* When working with multiple currencies, be aware of which currency your sObject uses so that you avoid inaccurate entries. The default currency matches the location at the organization level.
+
+=== Evaluate Values in a Salesforce Drop-down
+
+* If you are evaluating against a value in an existing drop-down list field in Salesforce, be sure to use the exact value in the dropdown. For example, if you use the value `US` to evaluate against the contents of a drop-down list that contains the value `USA`, the evaluation works, but you end up with two values in the dropdown: one for `US` and one for `USA`.
+
 === Fields To Null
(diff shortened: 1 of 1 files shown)


mauro-dentoni:
@alexandruatomei9, please, can you or anyone on the team help @george-hoffer reviewing this PR. Thanks in advance!

modules/ROOT/pages/salesforce/salesforce-connector-config-topics.adoc
```diff
@@ -313,50 +313,61 @@ output application/java
...
 
-=== Query
+In Mule 4, to achieve the same result:
 
-* Even though you can see the fields of an sObject and their corresponding types via DataSense, the `Query` operation returns all fields as `String`.
+. Call the *Get Updated Objects* operation, which has the `duration` parameter - exactly the same behavior as `initialTimeWindow` and returns the updated object IDs.
+. Call the *Retrieve* operation to specify the IDs obtained at step 1.
 
-* If you want to use the actual type of the field, you must convert that field to the desired type using a Transform (or Transform Message) component.
+The *Get Updated Objects* operation stores its results persistently by the timestamp of the latest objects.
```
alexandruatomei9, line 340: Starting with release 9.7.7(Mule 4) the description of GetUpdateObjects will change to this: 
`Retrieves the list of records that have been updated between the last time this method was called and the current server timestamp.
Salesforce API ignores the seconds portion of the specified dates so subsequent calls of this operation should take place at a rate of at least one minute, otherwise the operation will not return any results for the subsequent calls. `
 this is due to a bug fix that we implemented yesterday. There were some issues causing wither missing messages or sometimes duplicates, and they were caused in fact because salesforce does not care for the SECONDS or MILLISECONDS passed in the requests. So we have updated the operation to send requests with an interval of at leas one minute(if one request is sent at 12:14:05 and initialTimeWindow=3minutes we will in fact send to the api this interval: 12:11:00 - 12:14:00, after that the next request will be 12:14:00 - 12:15:00(if the operation is invoked after one minute))

george-hoffer:
Replacing this with a reference update.

PR #276, opened 2019-07-15
MULE-17104: Create WebSockets documentation.

emariotti3:


modules/ROOT/pages/websockets/websocket-about-websocket-connector.adoc
```diff
@@ -0,0 +1,60 @@
...
+== Why use Websockets
+
+WebSockets are a key piece of modern integration due to their ability to:
+
+* Seamlessly work on top of the existing HTTP infrastructure.
+* Avoid polling
+* Implement server push patterns
+* Establish bi-directional and full duplex communication between server and client.
+
+== Basic components
```
marianogonzalez: components -> concepts

modules/ROOT/pages/websockets/websocket-about-websocket-connector.adoc
```diff
@@ -0,0 +1,60 @@
...
+
+There are two basic socket types:
```
marianogonzalez: remove `basic`

modules/ROOT/pages/websockets/websocket-about-websocket-connector.adoc
```diff
@@ -0,0 +1,60 @@
...
+
+* Inbound: represents a websocket that will open upon receiving an incoming connection request from a client that supports websocket communication.
```
marianogonzalez: search and replaceAll: `websocket` -> `WebSocket`

marianogonzalez: Clarify that inbound sockets are initiated by external systems that connect through a WebSocket endpoint that the Mule application exposes and is listening on

modules/ROOT/pages/websockets/websocket-about-websocket-connector.adoc
```diff
@@ -0,0 +1,60 @@
...
+* Outbound: represents a websocket that will open a connection to other websockets through the 'connect' operation.
+
+All websockets, both inbound and outbound, will have their own unique socket ID to identify them.
+
+== Basic Operations Overview
+
+Using the WS or WSS protocol, the Websocket connector performs one of the following operations:
+
+* Inbound Listener/On Outbound Socket Message
```
marianogonzalez: this is a trigger, not an operation

modules/ROOT/pages/websockets/websocket-about-websocket-connector.adoc
```diff
@@ -0,0 +1,60 @@
...
++ Which will start executing a flow upon receiving an incoming message.
+* Send
++ Send message to a specific socket.
+* Broadcast
++ Send a given message to several sockets classified by path, type and group.
+* Open Outbound Socket
++ Open a connection to an external websocket.
+* On Inbound Connection
```
marianogonzalez: trigger

modules/ROOT/pages/websockets/websocket-about-websocket-connector.adoc
```diff
@@ -0,0 +1,60 @@
...
++ Start executing a flow when an external connection request is received and connection is successfully established.
```
marianogonzalez: `Start executing` -> Fires

modules/ROOT/pages/websockets/websocket-about-websocket-connector.adoc
```diff
@@ -0,0 +1,60 @@
...
+* Close
++ Close a connection established with another websocket.
+* On Socket Closed
```
marianogonzalez: trigger

modules/ROOT/pages/websockets/websocket-about-websocket-connector.adoc
```diff
@@ -0,0 +1,60 @@
...
++ Start executing a flow when a connection to a socket of any kind (INBOUND/OUTBOUND) has been closed.
+
+Websockets are built upon HTTP configuration and Listeners can be used to trigger flows in the same way as HTTP Listeners can.
```
marianogonzalez: I read this sentence 3 times and I can't figure out what it means.

modules/ROOT/pages/websockets/websocket-about-websocket-connector.adoc
```diff
@@ -0,0 +1,60 @@
...
+
+They even allow for both an HTTP and a Websocket Listener to be listening on the same port.
```
marianogonzalez: This is a feature of the connector, not of WebSockets itself. Also, it feels like this should be in the intro section

modules/ROOT/pages/websockets/websocket-about-websocket-connector.adoc
```diff
@@ -0,0 +1,60 @@
...
+
+While HTTP Connector provides only one Listener and one Requester which fulfill the Server and Client roles respectively, because Websockets allow full-duplex communication, we can establish communication using one Listener/Requester pair of type Outbound or one Listener/Requester pair of type Inbound. All mentioned operations are available for both types of sockets, we will distinguish them between client, server and connection management operations.
```
marianogonzalez: Wording is very confusing and seems to be missing the point. Are you trying to explain why there is one single config instead of being splitted in listener-config and requester-config like http? Is that something that should be explained here or is that a design consideration that only belongs in the spec?

modules/ROOT/pages/websockets/websocket-about-websocket-connector.adoc
```diff
@@ -0,0 +1,60 @@
...
+
+Some components, however, will be exclusive for Inbound Websockets or Outbound Websockets.
```
marianogonzalez: Delete this sentence

modules/ROOT/pages/websockets/websocket-connection-management-components.adoc
```diff
@@ -0,0 +1,65 @@
...
+ifndef::env-site,env-github[]
+include::_attributes.adoc[]
+endif::[]
+:keywords: connectors, ws, wss, websocket
+
+== About Connection Management Components
+
+These components provide tools for connection management for both server and client side operations. All of these operations are available for both Inbound and Outbound sockets.
+
+=== On Socket Closed
```
marianogonzalez: This is way too vague. This needs to explain what is it that you get in the payload and what in the attributes. Example has to be better

modules/ROOT/pages/websockets/websocket-connection-management-components.adoc
```diff
@@ -0,0 +1,65 @@
...
+----
+<flow name="onConnectionClosed">
+    <websocket:on-connection-closed ​path​=​"/ws/*" ​config-ref​=​"ws"​ />
+    <logger level="INFO" message="No more messages please: you shall not pass." />
+</flow>
+----
+
+Once this flow is invoked, no more messages can be processed because the socket has already been closed.
+
+=== Subscribe Groups
```
marianogonzalez: for this to be relevant to the reader, give a message segregation example use case.

modules/ROOT/pages/websockets/websocket-connection-management-components.adoc
```diff
@@ -0,0 +1,65 @@
...
+This operation allows a Websocket to subscribe to one or more groups. It can be invoked on the same socket as many times as necessary, each invokation being additive over the previous ones. Repeated groups will be ignored.
+
+[source,xml,linenums]
+----
+<flow name="subscribeToGroup">
+ <websocket:subscribe-groups ​socketId​=​"#[attributes.socketId]" groups​=​"#[['MARVEL', 'DC']]"​ />
+</flow>
+----
+
+This is especially useful for broadcast operations in which a message can be delivered to one or more sockets associated or subscribed to a specific group (or groups).
```
marianogonzalez: This should reference the openOutboundSocket operation and the InboundMessageListener which also allow to set the groups right when the connection is stablished.

I think that everything `groups` related is better served as a specific section about broadcasting and segregation in which these are explained in the context of that use case

modules/ROOT/pages/websockets/websocket-connection-management-components.adoc
```diff
@@ -0,0 +1,65 @@
...
+----
+
+=== Close Operation
+
+This operation enables a Websocket to close a currently open connection. The socket id whose connection is to be closed must be provided.
+
+[source,xml,linenums]
+----
+<flow name="closeConnection">
+    <websocket:close ​socketId​=​"#[attributes.socketId]"​ />
```
marianogonzalez: This has been renamed to `close-socket`

modules/ROOT/pages/websockets/websocket-connector-client-components.adoc
```diff
@@ -0,0 +1,115 @@
...
+=== Open Outbound Operation
+
+This operation allows an OUTBOUND socket to establish a connection to a websocket endpoint.
+
+In order to establish a connection a configuration reference must be provided. In addition, parameters like path and uri can be provided. For further reference check connector's technical reference.
+
+The following example opens a connection reusing most of the settings defined in the configuration reference, but others like the uri are overridden.
+
+[source,xml,linenums]
+----
```
marianogonzalez: explain what this returns... how do you get the socketId and why is that important. Show a meaningful example

modules/ROOT/pages/websockets/websocket-connector-client-components.adoc
```diff
@@ -0,0 +1,115 @@
...
+<flow ​name​=​"connect"​>
+    ​<websocket:open-outbound ​path​=​"/relative/path" uri​=​"ws://mulestocks.com/quotes"​ ​config-ref​=​"ws"​ />
```
marianogonzalez: this example is wrong. `uri` and `path` are exclusive parameters. They are meant for very different use cases. Explain those

modules/ROOT/pages/websockets/websocket-connector-client-components.adoc
```diff
@@ -0,0 +1,115 @@
...
+</flow>
+----
+
+==== Headers and Query Parameters
+
+Default headers and query parameters can be set at the configuration level if they will be used for every request.
```
marianogonzalez: `if they will be used for every request.`

* Looks like the `if` is a type
* This is misleading. Those default query parameters and headers will be sent when a new outbound connection is requested. They are not resent each time a message goes through the established socket.

modules/ROOT/pages/websockets/websocket-connector-client-components.adoc
```diff
@@ -0,0 +1,115 @@
...
+                        theme: 'USA'
+                    }
+                ]
+            ]]>
+        </websocket:query-params>
+    </websocket:open-outbound-socket>
+</flow>
+----
+
+==== Socket Id
```
marianogonzalez: This section is more about what the socket id is than how the operation uses it. This is a cornerstone concept. The The rules that the id must follow, are actually the rules of what an id is.

modules/ROOT/pages/websockets/websocket-connector-configuration.adoc
```diff
@@ -0,0 +1,75 @@
...
+ifndef::env-site,env-github[]
+include::_attributes.adoc[]
+endif::[]
+:keywords: connectors, ws, wss, websocket
+
+== About Websocket Connector Configuration
+
+There are two possible approaches to configuring a Websocket Connector. Both approaches will require an underlying HTTP Configuration from which several connection settings will be inherited, however these settings can be overridden in the Websocket configuration or at the operation level.
+
+=== Approach 1: Mimic HTTP Connector
```
marianogonzalez: This entire page is wrong. This was the spec decision process on why to pick one approach other the other. Only approach 2 got implemented.

This page should explain approach 2 in depth. Giving examples, explaining the parameters. For example:

* Explain that you can choose to only provide server settings if your app only accepts connections but doesn't open any
* Explain that you can choose to only provide client settings if your app only creates outbound connections but doesn't listen to incoming ones
* You can provide both if your app does both things
* Explain why you need to reference an http:listener-config in the server part, and how would you use that to configure TLS and other http related things
* explain how the default host, port, path parameter of the client settings section work and why are they optional and you might choose not to use them at all.
* Etc, etc, etc

modules/ROOT/pages/websockets/websocket-connector-server-components.adoc
```diff
@@ -0,0 +1,55 @@
...
+
+=== Inbound Listener
+
+The Inbound Listener component represents an websocket endpoint that clients can connect to. By hitting this endpoint, external clients not only establish a Websocket connection, they will trigger the owning server flow each time they send a message.
+
+[source,xml,linenums]
+----
+<flow name="onIncomingMessage">
+    <websocket:inbound-listener ​path​=​"/quotes"​ ​defaultGroups​=​"#[['AVENGERS', 'REVENGERS']]"​ ​config-ref​=​"ws"​ />
+    <set-variable name="myMessage" value="#[payload]" />
```
marianogonzalez: we advise customers not to put so many set-variable in a row. If you want to do this example, then use ee:transform.

That been said, this is not a very nice way of communicating this. You should describe the attributes object here and explain the use cases for each of its properties.

modules/ROOT/pages/websockets/websocket-connector-server-components.adoc
```diff
@@ -0,0 +1,55 @@
...
+    <set-variable name="senderSocketId" value="#[attributes.socketId]" />
+    <set-variable name="senderSocketType" value="#[attributes.socketType]" />
+    <set-variable name="senderSocketGroups" value="#[attributes.groups]" />
+</flow>
+----
+
+The payload will be set with the received message content, and its mime-type will be obtained from the incoming HTTP request header. If not set, the mime-type will default to one of:
+
+* application/octet-stream if it's binary content.
```
marianogonzalez: it's not about binary/text content. The underlying web socket protocol has the concept of frames. Frames are either binary or text. It's not the same thing

modules/ROOT/pages/websockets/websocket-connector-server-components.adoc
```diff
@@ -0,0 +1,55 @@
...
+* text/plain if it's text content.
+
+The output attributes will contain every attribute expected in the HttpRequestAttributes plus:
+
+    * Receiving socket's id.
+    * Receiving socket's type.
+    * Group list the receiving socket is currently subscribed to.
+
+=== On Inbound Connection
```
marianogonzalez: this fails to explain that if this flow fails, then the connection will be dropped, and that all messages will be queued until this flow ends.

It also doesn't explain that you cannot have two of these sources listening for the same path, nor it shows a useful usecase for this, like validating a header or updating some state.

modules/ROOT/pages/websockets/websocket-about-websocket-connector.adoc
```diff
@@ -0,0 +1,60 @@
...
+There are two basic socket types:
+
+* Inbound: represents a websocket that will open upon receiving an incoming connection request from a client that supports websocket communication.
+* Outbound: represents a websocket that will open a connection to other websockets through the 'connect' operation.
+
+All websockets, both inbound and outbound, will have their own unique socket ID to identify them.
+
+== Basic Operations Overview
+
+Using the WS or WSS protocol, the Websocket connector performs one of the following operations:
```
marianogonzalez: This is using the WS and WSS terms without explaining the concepts. At this point, the user has no idea that websocket URLs use the ws/wss schemes nor their relationship to TLS

modules/ROOT/pages/websockets/websocket-about-websocket-connector.adoc
```diff
@@ -0,0 +1,60 @@
+= About the Websockets Connector (Design Center)
```
marianogonzalez: this needs to be redone following the new connector documentation template

marianogonzalez (changes requested):

modules/ROOT/pages/websockets/websockets-connector-get-started.adoc
```diff
@@ -0,0 +1,133 @@
...
+and xref:websockets/websocket-connector-examples.adoc[Examples].
+
+=== Compatibility
+
+The WebSocket Connector is compatible with:
+
+|===
+|Application/Service|Version
+|Mule Runtime|4.2.2 or higher
+|Amazon Java SDK for EC2|v1.11.134
```
marianogonzalez: these JDKs are wrong. It also makes no sense to list JDKs at connector levels since all connectors work with whatever JDK the runtime supports

modules/ROOT/pages/websockets/websockets-connector-get-started.adoc
```diff
@@ -0,0 +1,133 @@
...
+|Java |1.8.0_x 
+|Anypoint Studio|7.4 or higher
+|===
+
+Audience without the "Additional Configuration Information" topic:
+
+== Audience
+
+* Starting user -- Read this Get Started document and create your Mule app in
```
marianogonzalez: this audience segregation doesn't apply here. This is a complex connector and all context is relevant.

modules/ROOT/pages/websockets/websockets-connector-get-started.adoc
```diff
@@ -0,0 +1,133 @@
...
+xref:websockets/websocket-connector-design-center.adoc[Design Center] or 
+xref:websockets/websocket-connector-studio.adoc[Anypoint Studio]. 
+* Power user - Read xref:websockets/websocket-connector-xml-maven.adoc[XML and Maven Support]
+and xref:websockets/websocket-connector-examples.adoc[Examples].
+
+
+
+== Next
+
+After you have met the prerequisites and experimented with templates
```
marianogonzalez: also doesn't make sense. There are not templates for this connector yet and this is pointing to articles already listed above as a next step

modules/ROOT/pages/websockets/websockets-connector-get-started.adoc
```diff
@@ -0,0 +1,133 @@
...
+== Next
+
+After you have met the prerequisites and experimented with templates 
+and examples, you are ready to create an app with
+xref:websockets/websocket-connector-design-center.adoc[Design Center] or 
+xref:websockets/websocket-connector-studio.adoc[Anypoint Studio].
+
+== See Also
+
+* https://forums.mulesoft.com[MuleSoft Forum]
```
marianogonzalez: this is missing links. A lot of articles in this doc are not reachable

marianogonzalez (changes requested):

modules/ROOT/pages/websockets/websockets-connector-broadcast.adoc
```diff
@@ -0,0 +1,44 @@
+= About Websocket Connection Management Components
+ifndef::env-site,env-github[]
+include::_attributes.adoc[]
+endif::[]
+:keywords: connectors, ws, wss, websocket, broadcast
+
+== About WebSockets Broadcasting
```
marianogonzalez: This page should be renamed and also include the docs for the send operation which is currently missing

marianogonzalez (changes requested):
@emariotti3 Rework the `See Next` links and make sure that all pages are linked from the landing page. Some of these pages are currently unreachable

modules/ROOT/pages/websockets/websockets-connector-messaging-operations.adoc
```diff
@@ -4,11 +4,33 @@ include::_attributes.adoc[]
 endif::[]
 :keywords: connectors, ws, wss, websocket, broadcast
 
-== About WebSockets Broadcasting
+== About WebSockets Operations
```
marianogonzalez: About WebSockets Operations -> About WebSockets Messaging Operations

modules/ROOT/pages/websockets/websockets-connector-messaging-operations.adoc
```diff
@@ -4,11 +4,33 @@ include::_attributes.adoc[]
...
+
+=== Send Operation
+
+The send operation allows the user to send a message through WebSocket identified by its socketId, once a connection has been established. This operation can be used from both ends of the connection, by Inbound and Outbound WebSockets.
```
marianogonzalez: can be used from both ends of the connection, by Inbound and Outbound WebSockets. -> can be used for both  Inbound and Outbound WebSockets.

modules/ROOT/pages/websockets/websockets-connector-messaging-operations.adoc
```diff
@@ -4,11 +4,33 @@ include::_attributes.adoc[]
...
+
+
+[source,xml,linenums]
+----
+    <flow name="sayHello">
+      <websocket:open-outbound-socket path="/chat/hello" config-ref="ws"/>
+      <websocket:send socketId="#[vars.socketId]" config-ref="ws">
```
marianogonzalez: vars.socketId -> attributes.socketId

marianogonzalez (changes requested):

marianogonzalez (approved):

george-hoffer:
Replaced by https://github.com/mulesoft/docs-connectors/pull/464

PR #295, opened 2019-07-26
DOCS-5405 Salesforce 9.7.7 update

george-hoffer:
See https://github.com/mulesoft/docs-connectors/pull/272#discussion_r304917706 for more info:
Starting with release 9.7.7(Mule 4) the description of GetUpdateObjects will change to this:
Retrieves the list of records that have been updated between the last time this method was called and the current server timestamp. Salesforce API ignores the seconds portion of the specified dates so subsequent calls of this operation should take place at a rate of at least one minute, otherwise the operation will not return any results for the subsequent calls. 
this is due to a bug fix that we implemented yesterday. There were some issues causing where missing messages or sometimes duplicates, and they were caused in fact because salesforce does not care for the SECONDS or MILLISECONDS passed in the requests. So we have updated the operation to send requests with an interval of at leas one minute. (if one request is sent at 12:14:05 and initialTimeWindow=3minutes we will in fact send to the api this interval: 12:11:00 - 12:14:00, after that the next request will be 12:14:00 - 12:15:00(if the operation is invoked after one minute))

### DIFF
--- a/modules/ROOT/pages/salesforce/salesforce-connector-reference-97.adoc
+++ b/modules/ROOT/pages/salesforce/salesforce-connector-reference-97.adoc
@@ -2220,27 +2220,27 @@ Retrieves the list of individual records that have been created or updated withi
 
 [[getUpdatedObjects]]
 === Get Updated Objects
+
 `<salesforce:get-updated-objects>`
 
+Retrieves a list of records that have been updated between the last time this method was called and the current server's timestamp. 
 
-Retrieves the list of records that have been updated between the last time this method was called and now. This method will save the timestamp of the latest date covered by Salesforce represented by `GetUpdatedResult#latestDateCovered`. 
-
-IMPORTANT: In order to use this method in a reliable way user must ensure that right after this method returns the result is stored in a persistent way since the timestamp of the latest. To reset the latest update time use `resetUpdatedObjectsTimestamp(String)`.
-
+The Salesforce API ignores the seconds portion of dates, so subsequent calls of this operation should take place at a rate of at least one minute, otherwise the operation does not return any results for the subsequent calls.
 
 ==== Parameters
-[%header%autowidth.spread]
+
+[%header,cols="20s,20a,35a,20a,5a"]
 |===
 | Name | Type | Description | Default Value | Required
 | Configuration | String | The name of the configuration to use. | | x
 | Type a| String |  Object type. The specified value must be a valid object for your organization. |  | x
-| Initial Time Window a| Number |  Time window (in minutes) used to calculate the start time (in time range) the first time this operation is called. For example, if initialTimeWindow equals 2, the start time is the current time (now) minus 2 minutes, then the range to retrieve the updated object is (now - 2 minutes; now). After first call the start time is calculated from the object store getting the last time this operation was exec |  | x
-| Fields a| Array of String |  The fields to retrieve for the updated objects |  | x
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/salesforce/salesforce-connector-reference-97.adoc
```diff
@@ -2220,27 +2220,29 @@ Retrieves the list of individual records that have been created or updated withi
...
 === Get Updated Objects
-`<salesforce:get-updated-objects>`
 
+`<salesforce:get-updated-objects>`
 
-Retrieves the list of records that have been updated between the last time this method was called and now. This method will save the timestamp of the latest date covered by Salesforce represented by `GetUpdatedResult#latestDateCovered`. 
+Retrieves a list of records that have been updated between the last time this method was called and the current server's timestamp. 
 
-IMPORTANT: In order to use this method in a reliable way user must ensure that right after this method returns the result is stored in a persistent way since the timestamp of the latest. To reset the latest update time use `resetUpdatedObjectsTimestamp(String)`.
+The Salesforce API ignores the seconds portion of dates, so subsequent calls of this operation should take place at a rate of at least one minute, otherwise the operation does not return any results for the subsequent calls.
```
dhirsch1: I suggest starting a new sentence with "Otherwise,..." The current sentence is long and hard to follow. If you don't want to do that, add a comma after "otherwise." 

Do you mean one call per minute? If so, "calls of this operation should take place at a rate of one per minute.

modules/ROOT/pages/salesforce/salesforce-connector-reference-97.adoc
```diff
@@ -2220,27 +2220,29 @@ Retrieves the list of individual records that have been created or updated withi
...
 
+For example, if a request is sent at `12:14:05` and the *Initial Time Window* field is set to 3 minutes, the *Get Updated Object* operation sends the Salesforce API this interval: `12:11:00 - 12:14:00`, after which the next request is `12:14:00 - 12:15:00` (if the operation is invoked after one minute).
```
dhirsch1: Again I split this into two sentences.

modules/ROOT/pages/salesforce/salesforce-connector-reference-97.adoc
```diff
@@ -2220,27 +2220,29 @@ Retrieves the list of individual records that have been created or updated withi
...
-| Fields a| Array of String |  The fields to retrieve for the updated objects |  | x
-| Update Headers a| <<RequestHeaders>> |  Salesforce Headers http://www.salesforce.com/us/developer/docs/api/Content/soap_headers.htm[More Info] |  | 
-| Target Variable a| String |  The name of a variable to store the operation's output. |  | 
-| Target Value a| String |  An expression to evaluate against the operation's output and store the expression outcome in the target variable. |  `#[payload]` | 
+| Initial Time Window a| Number |  Time window in minutes used to calculate the start time (in a time range) the first time this operation is called. For example, if *Initial Time Window* equals 2, the start time is the current time (now) minus 2 minutes, then the range to retrieve the updated object is (now - 2 minutes; now). After the first call, the start time is calculated from the object store getting the last time this operation was executed. |  | x
```
dhirsch1: This is hard to follow. Here are some suggestions:

Time window, in minutes, used...
If *Initial Time Window is 2,...

It would be better to remove the parentheses around "in a time range" and use commas instead.

I would add "and" before "the start time"

The last sentence is awkward. Maybe "the start time is retrieved from the object store as the last time this operation was executed. "

dhirsch1 (approved):
Approved but some of the sentences are long and awkward. I have suggestions to address this.

PR #300, opened 2019-07-29
fixed xrefs to anypoint-mq-new-major

hannanelson:


sleslie23 (approved):
I'm only approving the changes that fix broken xrefs. I did not review any other changes.

PR #314, opened 2019-08-06
DOCS-5682 NetSuite  v9.1 Connector Ref update

george-hoffer:


valkyrie69 (approved):
Assume due diligence was done (local build, all links tested)

PR #333, opened 2019-08-22
Add vm:consume to an example and fix nav file.

dhirsch1:
I have added a vm:consume operation to the example, but I know the example does not match the description because it has a Listener operation in it. Please let me know how to update the description or the example so they are both correct.

### DIFF
--- a/modules/ROOT/nav.adoc
+++ b/modules/ROOT/nav.adoc
@@ -259,7 +259,7 @@
 * xref:validation/validation-connector.adoc[Validation Module]
  ** xref:validation/validation-documentation.adoc[Validation Connector Reference]
 * xref:vm/vm-connector.adoc[VM Connector]
- ** xref:vm/vm-publish-listen.adoc[Example: Publish and Get a Response in the VM Connector]]
+ ** xref:vm/vm-publish-listen.adoc[Example: Publish and Listen to Messages with the VM Connector]]
  ** xref:vm/vm-dynamic-routing.adoc[Example: Dynamic Routing with the VM Connector]
  ** xref:vm/vm-publish-response.adoc[Example: Publish and Get a Response in the VM Connector]
  ** xref:vm/vm-publish-across-apps.adoc[Example: Send Messages across Different Apps]
--- a/modules/ROOT/pages/vm/vm-publish-across-apps.adoc
+++ b/modules/ROOT/pages/vm/vm-publish-across-apps.adoc
@@ -22,23 +22,13 @@ You can also send messages across apps by using Mule domains. Starting in Mule 4
              xmlns:vm="http://www.mulesoft.org/schema/mule/vm"
              xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
              xsi:schemaLocation="http://www.mulesoft.org/schema/mule/vm http://www.mulesoft.org/schema/mule/vm/current/mule-vm.xsd
-               http://www.mulesoft.org/schema/mule/domain http://www.mulesoft.org/schema/mule/domain/current/mule-domain.xsd">
+             http://www.mulesoft.org/schema/mule/domain http://www.mulesoft.org/schema/mule/domain/current/mule-domain.xsd">
 
-    <vm:config name="sharedVMConfig">
-      <vm:queue queueName="sharedQueue" queueType="PERSISTENT" />
-    </vm:config>
+  <vm:config name="sharedVMConfig">
+    <vm:queue queueName="sharedQueue" queueType="PERSISTENT" />
+  </vm:config>
 
 </mule-domain>
-----
-+
-. Place a listener in an application of the same domain:
-+
-[source,xml,linenums]
-----
-<flow name="crossAppListener">
-	<vm:listener queueName="sharedQueue" config-ref="sharedVMConfig" />
-	<logger />
-</flow>
 ----
 +
 . Publish messages from a second application, also of the same domain:
@@ -46,7 +36,7 @@ You can also send messages across apps by using Mule domains. Starting in Mule 4
 [source,xml,linenums]
 ----
 <flow name="crossAppRequester">
-	<vm:publish queueName="sharedQueue" config-ref="sharedVMConfig" />
+  <vm:publish queueName="sharedQueue" config-ref="sharedVMConfig" />
 </flow>
 ----
 +
@@ -55,9 +45,9 @@ You can also send messages across apps by using Mule domains. Starting in Mule 4
 [source,xml,linenums]
 ----
 <flow name="queueConsume">
-    <vm:consume queueName="myQueue" config-ref="sharedVMConfig" />
-    <logger />
+  <vm:consume queueName="myQueue" config-ref="sharedVMConfig" />
+  <logger />
 </flow>
 ----
 +
-In this example, the `listener`, `publish`, and `consume` operations point to the configuration defined in the domain.
+In this example, the `publish` and `consume` operations point to the configuration defined in mule-domain.
--- a/modules/ROOT/pages/vm/vm-publish-across-apps.adoc
+++ b/modules/ROOT/pages/vm/vm-publish-across-apps.adoc
@@ -19,10 +19,10 @@ You can also send messages across apps by using Mule domains. Starting in Mule 4
 [source,xml,linenums]
 ----
 <mule-domain xmlns="http://www.mulesoft.org/schema/mule/domain"
-             xmlns:vm="http://www.mulesoft.org/schema/mule/vm"
-             xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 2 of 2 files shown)


modules/ROOT/nav.adoc
```diff
@@ -259,7 +259,7 @@
 * xref:validation/validation-connector.adoc[Validation Module]
  ** xref:validation/validation-documentation.adoc[Validation Connector Reference]
 * xref:vm/vm-connector.adoc[VM Connector]
- ** xref:vm/vm-publish-listen.adoc[Example: Publish and Get a Response in the VM Connector]]
+ ** xref:vm/vm-publish-listen.adoc[Example: Publish and Listen to Messages with the VM Connector]]
```
ndinu, line 262: Why is this necessary? Which concerns did you have regarding the old tittle?

modules/ROOT/pages/vm/vm-publish-across-apps.adoc
```diff
@@ -33,24 +30,34 @@ So, you can create a VM configuration in a domain like in this:
...
 ----
 <flow name="crossAppRequester">
 	<vm:publish queueName="sharedQueue" config-ref="sharedVMConfig" />
 </flow>
 ----
-
-Notice how both the listener and publisher point to the same configuration defined in the domain.
\ No newline at end of file
++
+. Pull messages from the queue after they are published.
```
ndinu, line 91: I wouldn't say that this is another step, you either put a listener in the queue, or you consume them via the consume operation. this makes it seem like it would be valid to do both.

modules/ROOT/nav.adoc
dhirsch1: The old title there twice. So I changed the first occurrence to reflect the example the link to that example in the See Also section.(There is an error in the new ink that I need to correct.) Here is the duplicate occurrence: 
Screen Shot 2019-08-26 at 9 10 29 AM



Screen Shot 2019-08-26 at 9 11 40 AM

modules/ROOT/pages/vm/vm-publish-across-apps.adoc
dhirsch1: Please tell me how to get the example to be correct. Would the example be correct if I eliminate the listener code frlom it?

ndinu: You could say that instead of using a listener to process the messages you can use the consume operation to consume them.

modules/ROOT/nav.adoc
ndinu: Great 👍

modules/ROOT/pages/vm/vm-publish-across-apps.adoc
```diff
@@ -4,53 +4,96 @@ include::_attributes.adoc[]
 endif::[]
 :keywords: VM, queues, connector, publish, listen, response, domain, across applications
 
+The following examples show two ways to send messages across different apps:
 
+* Use a listener to receive the messages from the first app and a `publish` operation to feed the received messages to a second app.
+* Use a `publish` operation in a second app to pull messages from the queue of messages, and a `consume` operation to consume the messages.
```
ndinu: Use a `publish` operation in a second app to push messages to the queue of messages, and a `consume` operation to consume the messages.

ndinu (changes requested):

ndinu (approved):

modules/ROOT/pages/vm/vm-publish-across-apps.adoc
```diff
@@ -4,53 +4,96 @@ include::_attributes.adoc[]
...
 :keywords: VM, queues, connector, publish, listen, response, domain, across applications
 
+The following examples show two ways to send messages across different apps:
 
+* Use a listener to receive the messages from the first app and a `publish` operation to feed the received messages to a second app.
+* Use a `publish` operation in a second app to push messages to the queue of messages, and a `consume` operation to consume the messages.
 
+In both cases, the two apps must be in the same domain.
 
+== Publishing Messages to a Second App
```
george-hoffer: Salesforce style says to not use gerunds in headings

modules/ROOT/pages/vm/vm-publish-across-apps.adoc
```diff
@@ -4,53 +4,96 @@ include::_attributes.adoc[]
...
 
-Instead of using the `<vm:listener>` message source, you can consume messages in a VM queue through the `<vm:consume>` operation.
+To publish messages to a second app:
```
george-hoffer: This statement isn't needed -- just duplicates the heading. please remove.

modules/ROOT/pages/vm/vm-publish-across-apps.adoc
```diff
@@ -4,53 +4,96 @@ include::_attributes.adoc[]
...
 
-This operation can consume messages on demand, allowing you to:
+. Create a VM configuration in a domain, like this:
++
+[source,xml,linenums]
+----
+<mule-domain xmlns="http://www.mulesoft.org/schema/mule/domain"
+  xmlns:vm="http://www.mulesoft.org/schema/mule/vm"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://www.mulesoft.org/schema/mule/vm http://www.mulesoft.org/schema/mule/vm/current/mule-vm.xsd
```
george-hoffer: Please add a carriage return between "/vm http" so this line doesn't require users to scroll the code example.

modules/ROOT/pages/vm/vm-publish-across-apps.adoc
```diff
@@ -4,53 +4,96 @@ include::_attributes.adoc[]
...
+  http://www.mulesoft.org/schema/mule/domain http://www.mulesoft.org/schema/mule/domain/current/mule-domain.xsd">
```
george-hoffer: Same issue here -- forcing users to have a scroller for non-essential info causes angst

modules/ROOT/pages/vm/vm-publish-across-apps.adoc
```diff
@@ -4,53 +4,96 @@ include::_attributes.adoc[]
...
+. Publish messages from a second app in the same domain:
++
+[source,xml,linenums]
+----
+<flow name="crossAppRequester">
+  <vm:publish queueName="sharedQueue" config-ref="sharedVMConfig" />
+</flow>
+----
+
+In this example, the listener and `publish` operation point to the same queue (`sharedQueue`) and the same configuration (`sharedVMConfig`).
```
george-hoffer: Either add tick marks to listener or remove them from publish. And "operation" should be plural.

modules/ROOT/pages/vm/vm-publish-across-apps.adoc
```diff
@@ -4,53 +4,96 @@ include::_attributes.adoc[]
...
 [source,xml,linenums]
 ----
 <mule-domain xmlns="http://www.mulesoft.org/schema/mule/domain"
-             xmlns:vm="http://www.mulesoft.org/schema/mule/vm"
-             xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-             xsi:schemaLocation="http://www.mulesoft.org/schema/mule/vm http://www.mulesoft.org/schema/mule/vm/current/mule-vm.xsd
-               http://www.mulesoft.org/schema/mule/domain http://www.mulesoft.org/schema/mule/domain/current/mule-domain.xsd">
+  xmlns:vm="http://www.mulesoft.org/schema/mule/vm"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://www.mulesoft.org/schema/mule/vm http://www.mulesoft.org/schema/mule/vm/current/mule-vm.xsd
```
george-hoffer: Add carriage returns so example doesn't require scrolling right.

modules/ROOT/pages/vm/vm-publish-across-apps.adoc
```diff
@@ -4,53 +4,96 @@ include::_attributes.adoc[]
...
-	<vm:publish queueName="sharedQueue" config-ref="sharedVMConfig" />
+<flow name="queueConsume">
+  <vm:consume queueName="sharedQueue" config-ref="sharedVMConfig" />
+  <logger />
 </flow>
 ----
 
-Notice how both the listener and publisher point to the same configuration defined in the domain.
\ No newline at end of file
+In this example, the `publish` and `consume` operations point to the same queue (`sharedQueue`) and the same configuration (`sharedVMConfig`).
```
george-hoffer: Either init cap publish and consume or remove the tick marks. Make sure you check these operation names in Studio to be sure they exist and for capitalization.

george-hoffer (changes requested):

george-hoffer (approved):

PR #336, opened 2019-08-27
DOCS-5008 Fixed typo

valkyrie69:
And other minor edits.

### DIFF
--- a/modules/ROOT/pages/xml/xml-xslt.adoc
+++ b/modules/ROOT/pages/xml/xml-xslt.adoc
@@ -10,7 +10,6 @@ Although MuleSoft supports the XSLT standard, DataWeave is the recommended tool 
 
 This is an example XML input document:
 
-.Input Example
 [source,xml,linenums]
 ----
 <?xml version="1.0" encoding="UTF-8"?>
@@ -25,7 +24,6 @@ This is an example XML input document:
 
 You can transform the XML input document like this:
 
-.XSLT Transformation Script
 [source,xml,linenums]
 ----
 <xml-module:xslt-transform>
@@ -61,7 +59,6 @@ You can transform the XML input document like this:
 
 This script outputs the following document:
 
-.Output
 [source,xml,linenums]
 ----
 <table>
@@ -90,7 +87,6 @@ This script outputs the following document:
 
 Although this operation looks for the input document at the message payload level by default, you can supply your own input. For example, imagine that the cities are a JSON array inside the payload. You could then first transform the JSON to XML, and then run the XSL transformation:
 
-.XSLT Transformation Script
 [source,xml,linenums]
 ----
 <xml-module:xslt-transform>
--- a/modules/ROOT/pages/xml/xml-xslt.adoc
+++ b/modules/ROOT/pages/xml/xml-xslt.adoc
@@ -6,9 +6,9 @@ endif::[]
 
 The `<xml-module:xslt-transform>` element can take an input XML document and transform it using an XSL style sheet.
 
-MuleSoft supports the XSLT standard, but DataWeave is capable of achieving the same use cases and is the recommended tool for extracting and transforming XML documents.
+Although MuleSoft supports the XSLT standard, DataWeave is the recommended tool for extracting and transforming XML documents.
 
-This is an example input XML document:
+This is an example XML input document:
 
 .Input Example
 [source,xml,linenums]
@@ -59,7 +59,7 @@ You can transform the XML input document like this:
 </xml-module:xslt-transform>
 ----
 
-The script above outputs the following document:
+This script outputs the following document:
 
 .Output
 [source,xml,linenums]
@@ -88,7 +88,7 @@ The script above outputs the following document:
 </table>
 ----
 
-Although by default this operation looks for the input document at the message payload level, you can supply your own input. For example, imagine that the cities were a JSON array inside the payload. You could first transform the JSON to XML, and then run the XSL transformation. For example:
+Although this operation looks for the input document at the message payload level by default, you can supply your own input. For example, imagine that the cities are a JSON array inside the payload. You could then first transform the JSON to XML, and then run the XSL transformation:
 
 .XSLT Transformation Script
 [source,xml,linenums]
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/xml/xml-xslt.adoc
```diff
@@ -6,9 +6,9 @@ endif::[]
 
 The `<xml-module:xslt-transform>` element can take an input XML document and transform it using an XSL style sheet.
 
-Note that though MuleSoft supports the XSLT standard, DataWeave is capable of achieving the same use cases and is the recommended tool for extracting and transforming XML documents.
+MuleSoft supports the XSLT standard, but DataWeave is capable of achieving the same use cases and is the recommended tool for extracting and transforming XML documents.
```
kbostrom: Another way of saying this that puts the "but" first is "Although MuleSoft supports, DataWeave can do it too."
You might want that emphasis; or, you might not! Your decision; both are fine.

kbostrom: is capable of > can

modules/ROOT/pages/xml/xml-xslt.adoc
```diff
@@ -6,9 +6,9 @@ endif::[]
...
 
-Consider the following input document:
+This is an example input XML document:
```
kbostrom: Forgive my ignorance: can you review and be sure that all these modifiers are in the right position:

     example > input > XML > document?

Is it a document serving as an example of XML input? Is it an XML document that shows example input? Or some other combination? I could not parse the string with any certainty, but your audience might be just fine with it.

kbostrom: Ah! I see below that you use "the XML input document," so I think what you want is "example XML input document" above.

modules/ROOT/pages/xml/xml-xslt.adoc
```diff
@@ -59,7 +59,7 @@ You can transform it like this:
 </xml-module:xslt-transform>
 ----
 
-The script above will output the following document:
+The script above outputs the following document:
```
kbostrom: The script above > This script

modules/ROOT/pages/xml/xml-xslt.adoc
```diff
@@ -88,7 +88,7 @@ The script above will output the following document:
 </table>
 ----
 
-Although by default this operation looks for the input document at the message payload level, you can supply your own input. For example, imagine that the cities were a JSON array inside the payload. You could first transform it to XML before  running the XSL transformation, for example:
+Although by default this operation looks for the input document at the message payload level, you can supply your own input. For example, imagine that the cities were a JSON array inside the payload. You could first transform the JSON to XML, and then run the XSL transformation. For example:
```
kbostrom:
```suggestion
Although this operation looks for the input document at the message payload level by default, you can supply your own input. For example, imagine that the cities are a JSON array inside the payload. You could then first transform the JSON to XML, and then run the XSL transformation:
```

modules/ROOT/pages/xml/xml-xslt.adoc
```diff
@@ -131,11 +131,11 @@ In a more complex use case, the script you use might depend on some external con
 </flow>
 ----
 
-. This flow is triggered through an HTTP request.
-. Assuming that the `userId` is provided as a query parameter, the example uses the File connector to read the file that contains the correct XSL and store it in a variable.
-. The example executes a transformation using an expression to point to the XSL sheet obtained through the Read operation.
+. The example flow is triggered through an HTTP request.
+. The example assumes that the `userId` is provided as a query parameter and uses the File connector to read the file that contains the correct XSL and stores it in a variable.
```
kbostrom:
```suggestion
. The example is based on the `userId` value being provided as a query parameter and the File connector being used to read the file that contains the correct XSL and store it in a variable.
```

kbostrom: Is File the name of the connector, or is it a type of connector in general?
If the former then...the short official name is File Connector (?) 
If the latter, lowercase: the file connector

modules/ROOT/pages/xml/xml-xslt.adoc
```diff
@@ -131,11 +131,11 @@ In a more complex use case, the script you use might depend on some external con
...
+. The example then executes a transformation using an expression to point to the XSL sheet obtained through the `Read` operation.
 
-Note that the HTTP and File connectors are simply used as examples. They are not required to perform similar use cases.
+The HTTP and File connectors are used as examples and are not required to perform similar use cases.
```
kbostrom: ditto re. connector name vs. type of connector

modules/ROOT/pages/xml/xml-xslt.adoc
```diff
@@ -161,12 +161,12 @@ The XML module, like the XSL language, provides context properties that can pass
 
 . The XSL style sheet declares a `param` called `vtn`.
 . The style sheet references the assigned value of the `param` using the `$` prefix.
-. The example use the `context-properties` parameter to pass the values in.
+. The example uses the `context-properties` parameter to pass the values in.
```
kbostrom: "to pass the values in" > "to pass in the values" or "to import the values"

(The verb is "to pass in.")

modules/ROOT/pages/xml/xml-xslt.adoc
```diff
@@ -161,12 +161,12 @@ The XML module, like the XSL language, provides context properties that can pass
...
 
 
-== Output Mime Type
+== Output MIME Type
 
-The most common use case for XSLT is to output a new XML document. However, XSLT can produce documents in other formats, such as JSON, CSV, or simply a text or number value. Because of this, the produced value will not have any specific mime type assigned. It is recommended that you specify one using the `outputMimeType` parameter.
+The most common use case for XSLT is to output a new XML document. However, XSLT can produce documents in other formats, such as JSON, CSV, or a text or number value, which means that the produced value will not have any specific MIME type assigned. You should specify a MIME type using the `outputMimeType` parameter.
```
kbostrom:
```suggestion
The most common use case for XSLT is to output a new XML document. However, XSLT can produce documents in other formats, such as JSON, CSV, or a text or number value, which means that the produced value does not have any specific MIME type assigned. You should specify a MIME type using the `outputMimeType` parameter.
```

kbostrom (approved):
Here you go1

PR #337, opened 2019-08-28
Update sap-connector.adoc

FilipBogdanSV:


george-hoffer (approved):

pengjasonsalesforce:
should we also add that for rtf?

PR #341, opened 2019-09-05
Workday Connector Doc Templated - Add files via upload DO NOT MERGE

janbily:
- workday-connector-config-topics.adocs is not uploaded as there are not additional configuration information for this connector (at least I am not aware of any)


valkyrie69 (approved):
Approved with the assumption that local build was done and all links verified.

PR #352, opened 2019-09-12
DOCS-4284

valkyrie69:
change max buffer size to max in memory size and other cleanup

### DIFF
--- a/modules/ROOT/pages/file/file-documentation.adoc
+++ b/modules/ROOT/pages/file/file-documentation.adoc
@@ -7,28 +7,24 @@ Support Category: Select
 
 File Connector v1.3
 
-The File connector manipulates file systems mounted on a host operation system.
+The Anypoint Connector for File (File Connector) manipulates file systems mounted on a host operation system.
 
-This class serves as both extension definition and configuration. Operations are based on the standard BaseFileSystemOperations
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/file/file-documentation.adoc
```diff
@@ -80,110 +75,104 @@ A ConnectionProvider which provides instances of FileSystem from instances of Fi
 
 `<file:copy>`
 
-Copies the file at the sourcePath into the targetPath.
-
-If targetPath doesn't exist, and neither does its parent, then an attempt is made to create depending on the value of the createParentFolder argument. If such an argument occurs, then a FILE:ILLEGAL_PATH is thrown.
+Copies the file or directory specified in *Source Path* into the *Target Path*. The source path can be either a file or a directory. If it points to a directory, then it is copied recursively.
```
facunlk, line 78: If the argument createParentFolder is not provided and the path doesn't exist, it will throw a FILE:ILLEGAL_PATH error. This should be documented: https://github.com/mulesoft/mule-file-connector/blob/c1f72b136e999384533cecbba35b843b183ef525/src/main/java/org/mule/extension/file/internal/FileOperations.java#L167

facunlk (changes requested):

facunlk (approved):

sarathecoubian (approved):

PR #397, opened 2019-10-09
Salesforce Composite 2.3.0 Connector Reference

dancazacu1:


george-hoffer (approved):

george-hoffer:
This doc appears updated. See https://docs.mulesoft.com/connectors/salesforce/salesforce-composite-connector-reference. Also the repo now lists as unknown.

PR #398, opened 2019-10-10
X12 EDI - Updated the Note about size restriction

amitfernandes:


modules/ROOT/pages/x12-edi/x12-edi-connector.adoc
```diff
@@ -22,8 +22,7 @@ release notes is listed in the See Also section of this document.
 To use the X12 EDI connector in a production environment, you must
 have purchased a MuleSoft license for Anypoint B2B.
 
-NOTE: X12 EDI Connector supports files up to 20MB. If you process files larger than 20MB, consider that the connector has a 40x memory requirement that corresponds to the EDI file size. 
-For example, a 30MB X12 (or `EDIFACT`) file needs 1.4GB of memory, so it's important to consider this memory requirement in conjunction with your TPS needs for large files.
+NOTE: The X12 EDI Connector supports files up to 15MB, and the memory requirement is roughly 40 to 1. For example, a 1MB file requires up to 40MB of memory to process, so it's important to consider this memory requirement in conjunction with your TPS needs for large X12 files. This is not an exact figure as the value may vary according to the complexity of the mapping instructions.
```
kbostrom: NOTE: EDIFACT EDI Connector supports files of up to 15 MB, and the memory requirement is roughly 40 to 1. For example, a 1-MB file requires up to 40 MB of memory to process, so it's important to consider this memory requirement in conjunction with your TPS needs for large EDIFACT files. This is not an exact figure; the value might vary according to the complexity of the mapping instructions.

kbostrom: (I just copied this in from the previous edit.)

kbostrom (approved):
Done!

PR #399, opened 2019-10-10
EDIFACT EDI - Added a note about File Size restrictions

amitfernandes:


modules/ROOT/pages/edifact/edifact-edi-connector.adoc
```diff
@@ -19,7 +19,7 @@ release notes is listed in the See Also section of this document.
 To use the EDIFACT EDI connector in a production environment, you must
 have purchased a MuleSoft license for Anypoint B2B.
 
-NOTE: The EDIFACT EDI Connector supports files up to 15MB. The EDIFACT EDI connector's memory requirement is roughly 40 to 1. For example a 1MB file will require up to 40MB of memory to process. This is not an exact figure; the value may vary according to the complexity of the mapping instructions. It is important to consider this memory requirement in conjunction with your TPS needs for large EDIFACT files.
+NOTE: The EDIFACT EDI Connector supports files up to 15MB, and the memory requirement is roughly 40 to 1. For example, a 1MB file requires up to 40MB of memory to process, so it's important to consider this memory requirement in conjunction with your TPS needs for large EDIFACT files. This is not an exact figure as the value may vary according to the complexity of the mapping instructions.
```
dhirsch1: I am not sure if "as" translates well. I suggest using "because" instead.

dhirsch1 (approved):
Approved with one comment.

modules/ROOT/pages/edifact/edifact-edi-connector.adoc
kbostrom: This is a good catch, Diane, but because the main point is that the figure is inexact, I suggest using a semicolon instead, to give the first part equal, if not primary, importance.

modules/ROOT/pages/edifact/edifact-edi-connector.adoc
```diff
@@ -19,6 +19,8 @@ release notes is listed in the See Also section of this document.
 To use the EDIFACT EDI connector in a production environment, you must
 have purchased a MuleSoft license for Anypoint B2B.
 
+NOTE: The EDIFACT EDI Connector supports files up to 15MB, and the memory requirement is roughly 40 to 1. For example, a 1MB file requires up to 40MB of memory to process, so it's important to consider this memory requirement in conjunction with your TPS needs for large EDIFACT files. This is not an exact figure as the value may vary according to the complexity of the mapping instructions.
```
kbostrom:
```suggestion
NOTE: EDIFACT EDI Connector supports files of up to 15 MB, and the memory requirement is roughly 40 to 1. For example, a 1-MB file requires up to 40 MB of memory to process, so it's important to consider this memory requirement in conjunction with your TPS needs for large EDIFACT files. This is not an exact figure; the value might vary according to the complexity of the mapping instructions.
```

kbostrom (approved):
Done!

modules/ROOT/pages/edifact/edifact-edi-connector.adoc
mcherem: Thanks Kim. And thanks Diane.

PR #416, opened 2019-10-18
Salesforce B2C CC - (Replaced by PR 441)

george-hoffer:


george-hoffer:
Replaced by https://github.com/mulesoft/docs-connectors/pull/441

PR #407, opened 2019-10-15
Update as2-connector.adoc

mcherem:
Adding alias to the AS2 connector to eliminate orphan  link.

PR #420, opened 2019-10-19
Update page-aliases for as2-connector

fermujica:
Based on this conversation: https://github.com/mulesoft/docs-connectors/pull/407
We need to update this page-aliases


mcherem (approved):
Peer review.

PR #429, opened 2019-10-25
DOCS-6336 fixes

george-hoffer:


george-hoffer:
Replaced by PR https://github.com/mulesoft/docs-connectors/pull/444

PR #439, opened 2019-10-30
First commit

dhirsch1:
Please Read:

I split some of the concepts out but the Advantages of Full Duplex piece does not seem to fit in the separate concepts topic. Also, it seems weird to split out things like the WebSocket ID and Groups because they are just as integral as triggers and operations. Therefore, I don't think the current split works. Do you have any suggestions?


dhirsch1:
This PR is no longer relevant.

PR #447, opened 2019-11-05
Update x12-edi-connector.adoc

dsosnoski:
write-configuration
Add description of useSuppliedValues configuration parameter added in 2.1.0 release. The attached image updates that used in the document under "About the Writer Tab"


mcherem (approved):
Doc review.

PR #457, opened 2019-11-11
DOCS-6231 update

george-hoffer:
https://www.mulesoft.org/jira/browse/DOCS-6231


george-hoffer:
Not going to work. Back to the drawing board.

PR #460, opened 2019-11-12
deleting broken images

valkyrie69:


mcherem (approved):
Peer review - Looks like an interim fix - assuming this is built and tested locally.

PR #472, opened 2019-11-21
WebSockets bug fixes

george-hoffer:
DOCS-6634

### DIFF
--- a/modules/ROOT/pages/websockets/websockets-connector-server-components.adoc
+++ b/modules/ROOT/pages/websockets/websockets-connector-server-components.adoc
@@ -42,16 +42,16 @@ Clients can also indicate the default groups that this endpoint automatically su
 
 == Send a Response
 
-Unlike the `<http:listener>` element which automatically sends a response when the flow ends, this component  never automatically sends a response. This is because the web sockets protocol is full duplex and therefore provides no guarantee that you actually want or need to send a response at all. Depending on your use case, you might want to simply process the messages without sending a response, or only send responses on certain cases.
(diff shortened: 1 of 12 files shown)


modules/ROOT/pages/websockets/websockets-connector-server-components.adoc
```diff
@@ -42,16 +42,16 @@ Clients can also indicate the default groups that this endpoint automatically su
 
 == Send a Response
 
-Unlike the `<http:listener>` element which automatically sends a response when the flow ends, this component  never automatically sends a response. This is because the web sockets protocol is full duplex and therefore provides no guarantee that you actually want or need to send a response at all. Depending on your use case, you might want to simply process the messages without sending a response, or only send responses on certain cases.
+Unlike the `<http:listener>` element that automatically sends a response when a flow ends, this component  never automatically sends a response. This is because the web sockets protocol is full duplex and therefore provides no guarantee that you need to or want to send a response. Depending on your use case, you might want to process the messages without sending a response, or only send responses on certain cases.
```
valkyrie69:
```suggestion
Unlike the `<http:listener>` element that automatically sends a response when a flow ends, this component  never automatically sends a response. This is because the WebSocket protocol is full duplex and therefore provides no guarantee that you need to or want to send a response. Depending on your use case, you might want to process the messages without sending a response, or send responses only on certain cases.
```

valkyrie69 (approved):
I'm pretty sure that all cases of web socket or web sockets ought to be WebSocket.  In fact, Mariano made a comment on the original PR to change them all to WebSocket so on my initial pass of these files I changed all these web socket instances to WebSocket. :-)

PR #479, opened 2019-11-22
WIP DO NOT EDIT- OUTDATED SF marketing cloud user dh - Release 2.0

dhirsch1:
The User Guide needed extensive rewording. I did that, but it needs to be re-checked for accuracy. I do not fully understand the Providing a Subclass as a Type to a Complex Field so I did not reword that section as much as the others.  We will get this out quickly, but our editor will have to look at this at a later point in time.


valkyrie69:
Way outdated and hopelessly conflicted.

PR #499, opened 2019-12-04
MS Dynamics 365 M4 -- Style and link update

george-hoffer:


kbostrom (approved):
Thanks for waiting!
I reviewed only the changed text.

PR #500, opened 2019-12-04
DOCS-6688 Fix garbled text

valkyrie69:


modules/ROOT/pages/aggregator/aggregator-examples.adoc
```diff
@@ -23,7 +23,7 @@ So, you could use this:
 </aggregators:size-based-aggregator>
```
kbostrom, line 23: INSTRUCTIONS FOR EDIT

Hi Kim, Can you please edit this PR (only the green text--the scope is a bug fix, not a rewrite). When you get to it is fine--not urgent. Thank you!

modules/ROOT/pages/aggregator/aggregator-examples.adoc
```diff
@@ -23,7 +23,7 @@ So, you could use this:
...
 ----
 
-If you want to execute some action every time your list of names is filled by a 10%. You do not want to do this for every new element because that would cause too many executions. You can assume that the logic to be executed is in another flow named `partialAction`. We can you the `incremental-aggregation` route for this.
+If you do not want to execute some action for every new element because that would cause too many executions, you can instead use `incremental-aggregation`. For example, you could execute an action every time your list of names increases by an increment of 10% as shown in the below example. You can assume that the logic to be executed is in another flow named `partialAction`.
```
kbostrom: Questions about "execute some action":

Did you intend "execute" or "repeat"?
Did you mean "some actions"? "an action"? " "a particular action"?

kbostrom: Also, can the final sentence follow the example, so that we can introduce the example directly?
```suggestion
If you do not want to execute an action for every new element because that would cause too many executions, you can use `incremental-aggregation`. For example, you could execute an action every time your list of names increases by 10%, as shown here:

You can assume that the logic to be executed is in another flow named `partialAction`. 
```

modules/ROOT/pages/aggregator/aggregator-examples.adoc
```diff
@@ -42,7 +42,7 @@ If you want to execute some action every time your list of names is filled by a
 </aggregators:size-based-aggregator>
 ----
 
-But what if you do not want to wait forever to fill every batch. In that case, you might want to process all the events that you have even if the `maxSize` is not reached. For this, you can configure a timeout for your aggregator. This example sets the waiting limit to 10 seconds.
+But what if you do not want to wait forever to fill every batch. In that case, you might want to process all the events that you have even if the `maxSize` is not reached. For this, you can configure a timeout for your aggregator. The below example sets the waiting limit to 10 seconds.
```
kbostrom:
```suggestion
If you do not want to wait the time it takes to fill every batch, you can process all the events that you have, even if the `maxSize` value is not reached, by setting a timeout for your aggregator. The following example sets the timeout to 10 seconds:
```

modules/ROOT/pages/aggregator/aggregator-examples.adoc
```diff
@@ -74,9 +74,9 @@ To make sure that the logic that processes all aggregated elements executes once
 ----
 
 For this case, the data types that come from the `aggregation-complete` route and `aggregator-listener` are the same. So, for the execution of the flow `whatToDoWithAllTheseNames`, it does not matter where the data comes from.
-However, with this configuration, in the case the aggregation is complete, the flow `whatToDoWithAllTheseNames` will trigger twice, once from the `aggregation-complete` route `flow-ref` and again from the `flow-ref` in the flow with the `aggregator-listener`.
+However, with this configuration, in the case the aggregation is complete, the flow `whatToDoWithAllTheseNames` will trigger twice: once from the `aggregation-complete` route `flow-ref`, and again from the `flow-ref` in the flow with the `aggregator-listener`.
```
kbostrom:
```suggestion
However, with this configuration, when the aggregation is complete, the flow `whatToDoWithAllTheseNames` triggers twice: once from the `aggregation-complete` route `flow-ref`, and again from the `flow-ref` in the flow with the `aggregator-listener`.
```

modules/ROOT/pages/aggregator/aggregator-examples.adoc
```diff
@@ -74,9 +74,9 @@ To make sure that the logic that processes all aggregated elements executes once
...
 
-To avoid multiple executions of the same logic, the `flow-ref` after the `aggregator-listener` should be wrapped in a `choice` component that uses the `isAggregationComplete` attribute to check for an incomplete aggregation. That way, the `flow-ref` is only called when a timeout triggered that flow.
+To avoid multiple executions of the same logic, the `flow-ref` after the `aggregator-listener` should be wrapped in a `choice` component that uses the `isAggregationComplete` attribute to check for an incomplete aggregation. That way, the `flow-ref` is called only when a timeout triggered that flow.
```
kbostrom: Again suggest splitting para text around the example:
```suggestion
To avoid multiple executions of the same logic, wrap the `flow-ref` value `aggregator-listener` in a `choice` component that uses the `isAggregationComplete` attribute to check for an incomplete aggregation:

Using this syntax, `flow-ref` is called only when a timeout triggers that flow.
```

modules/ROOT/pages/aggregator/aggregator-examples.adoc
```diff
@@ -146,7 +146,7 @@ The flow also needs a listener for timeouts:
 <flow-ref name="whatToDoWithAllTheseNamesFromSomeCountry"/>
 ----
 
-Notice that the example does not configure an eviction time in the aggregator. That is because the default value is `0` (Evict right after the group is released), which is the desired behavior. So if a group completed or timed out, a subsequent value can be added to the group without being rejected.
+Notice that the example does not configure an eviction time in the aggregator. That is because the default value is `0`, which means the eviction takes place right after the group is released, which is the desired behavior. So if a group completed or timed out, a subsequent value can be added to the group without being rejected.
```
kbostrom: Is "eviction time" = "timeout"? If so, choose just one term and use it throughout, to avoid confusion.

kbostrom: For nonnative readers, "takes place" > "occurs"
```suggestion
Notice that the example does not configure an eviction time in the aggregator. That is because the default value is `0`, which means the eviction occurs directly after the group is released, which is the desired behavior. Therefore, if a group finishes or is timed out, a subsequent value can be added to the group without being rejected.
```

kbostrom (approved):
Here you go!

PR #526, opened 2019-12-19
AMQP update.

fsgonz:


modules/ROOT/pages/amqp/amqp-connector.adoc
```diff
@@ -6,16 +6,16 @@ endif::[]
...
 Support Category: https://www.mulesoft.com/legal/versioning-back-support-policy#anypoint-connectors[Select]
 
-Anypoint Connector for AMQP (AMQP Connector) v1.4
+Anypoint Connector for AMQP (AMQP Connector) v1.5
 
 Release Notes: xref:release-notes::connector/connector-amqp.adoc[AMQP Connector Release Notes] +
 Exchange: https://www.mulesoft.com/exchange/com.mulesoft.connectors/mule-amqp-connector/[AMQP Connector]
 
-AMQP Connector enables your application to do messaging using an AMQP 0.9.1 compliant broker. The connector's main features include:
+AMQP Connector enables your application to publish and consume messages using an AMQP 0.9.1 compliant broker. The connector's main features include:
```
kbostrom: hyphen:
```suggestion
AMQP Connector enables your application to publish and consume messages using an AMQP 0.9.1-compliant broker. The connector's main features include:
```

modules/ROOT/pages/amqp/amqp-connector.adoc
```diff
@@ -6,16 +6,16 @@ endif::[]
```
kbostrom, line 6: Instructions for edit review:

Green only and do not edit the file named amqp-documentation.adoc (that is the generated reference file). By EOD would be great, but I will understand if you can't turn it around on such short notice, so then just do it as soon as you can.

modules/ROOT/pages/amqp/amqp-connector.adoc
```diff
@@ -25,14 +25,14 @@ AMQP Connector enables your application to do messaging using an AMQP 0.9.1 comp
 * Prefetch size and count "quality of service" settings
 * noLocal and exclusive consumers
 * TLS/SSL connectivity
-* Message Builder functionality similar to the one provided by the JMS Connector
+* Message builder functionality similar to the one provided by the JMS Connector
```
kbostrom:
```suggestion
* Message builder functionality similar to that provided by JMS Connector
```

modules/ROOT/pages/amqp/amqp-connector.adoc
```diff
@@ -25,14 +25,14 @@ AMQP Connector enables your application to do messaging using an AMQP 0.9.1 comp
...
 
 [[configuration_settings]]
 == Configure the Connector
 
-Out of the box, the AMQP connector comes with a fine tuned set of default values, both for publishing and consuming messages. This means that the only requirement is for you to configure which connection should be used.
+Out of the box, the AMQP connector comes with a fine tuned set of default values, both for publishing and consuming messages. This means that the only requirement is for you to configure which connection to use.
```
kbostrom:
```suggestion
AMQP Connector comes out of the box with a fine-tuned set of default values for both publishing and consuming messages. This means that the only requirement is for you to configure which connection to use. 
```

modules/ROOT/pages/amqp/amqp-connector.adoc
```diff
@@ -62,9 +62,9 @@ Anypoint Exchange and click *Dependency Snippets*.
 
 === Define Global Defaults
 
-The AMQP `config` allows you to define multiple parameters that will be used by default while consuming or publishing messages. This way, you can define a global default behavior for all the operations associated to the config and then override each parameter only in the operations that require a custom behavior.
+The AMQP `config` allows you to define multiple parameters to use by default while consuming or publishing messages. This way, you can define a global default behavior for all the operations associated with the config and then override each parameter only in the operations that require a custom behavior.
```
kbostrom: What should we call `config`? Is it a command? a value? some other thing?
"The... `config` <option> enables you to...."

Also, we later refer to "config" without the syntax formatting. In that case, should we be spelling out "configuration" because that's what we mean: the thing rather than the syntax?

kbostrom: I had to think about the meaning of this for a while, and I think I got tripped up by the seeming contradiction between "multiple" and "default" (a default typically implying one preferred "thing"). 
I could very well be overthinking this, but I wonder if we should call this "a set of default parameters"?

modules/ROOT/pages/amqp/amqp-connector.adoc
```diff
@@ -76,9 +76,9 @@ For example, we can define a prefetch size window in terms of size and in terms
 </amqp:config>
 ----
 
-In this case, every `consume` operation or `listener` associated to this config uses an acknowledge mode of type `IMMEDIATE`, while every time we do a `publish` or `publish-consume`, the outgoing Message requests a confirm from the broker.
+In the above example, every `consume` operation or `listener` associated to this config uses an acknowledge mode of the type `IMMEDIATE`, while every time we do a `publish` or `publish-consume`, the outgoing message requests a confirm from the broker.
```
kbostrom: What is the object for the syntax `listener` (like "operation" is the object for `consume`)?

Also, ditto use of "config" vs. `config` or "configuration"?

kbostrom:
```suggestion
In the previous example, every `consume` operation or `listener` associated with this config uses an acknowledgment mode of the type `IMMEDIATE`, while for every `publish` or `publish-consume` operation, the outgoing message requests a confirmation from the broker.
```
Please adapt as necessary to ensure tech accuracy.

modules/ROOT/pages/amqp/amqp-connector.adoc
```diff
@@ -76,9 +76,9 @@ For example, we can define a prefetch size window in terms of size and in terms
...
 
-All of this properties can be overridden at the operation level when required. Let's say we need to `publish` a message that does not require confirms from the broker:
+You can override all of these properties at the operation level when required. For example, if you need to `publish` a message that does not require confirmation from the broker:
```
kbostrom:
```suggestion
You can override all of these properties at the operation level when required. For example, if you need to publish a message that does not require confirmation from the broker:
```

modules/ROOT/pages/amqp/amqp-consume.adoc
```diff
@@ -5,44 +5,44 @@ endif::[]
 :imagesdir: ../../assets/images/
 
 
-The Consume operation in the AMQP connector provides the ability to consume a message at any given time of the flow, from any given AMQP queue.
+The `consume` operation in AMQP Connector provides the ability to consume a message at any given time of the flow, from any given AMQP queue.
```
kbostrom:
```suggestion
The `consume` operation in AMQP Connector provides the ability to consume a message at any given time in the flow, from any given AMQP queue.
```

modules/ROOT/pages/amqp/amqp-consume.adoc
```diff
@@ -5,44 +5,44 @@ endif::[]
...
+
 The syntax to consume a message from a queue is:
 
 [source,xml,linenums]
 ----
 <amqp:consume config-ref="AMQP_config" queueName="#[vars.queue]"/>
 ----
 
-This operation consumes the first available message in the queue identified by the `queueName`, and converts it to a `AmqpMessage` resulting in the following structure:
+This operation consumes the first available message in the queue identified by the `queueName`, and converts it to an `AmqpMessage`, which results in the following structure:
```
kbostrom:
```suggestion
This operation consumes the first available message in the queue identified by the `queueName` value, and converts it to an `AmqpMessage` value, which results in the following structure:
```

modules/ROOT/pages/amqp/amqp-consume.adoc
```diff
@@ -5,44 +5,44 @@ endif::[]
...
 
 * The message's content as payload
 * The message's metadata in the message attributes
 
-Once received, the message will be immediately ACKed by default. If you want to control the ACK of the message after some processing, then `ackMode` should be set to `MANUAL`
-For more information regarding a message ACK, see xref:amqp/amqp-ack.adoc[How to handle Acknowledgment].
+Once received, the message is immediately ACKed by default. If you want to control the ACK of the message after some processing, then set the `ackMode` to `MANUAL`.
```
kbostrom:
```suggestion
After it is received, the message is immediately acknowledged by default. If you want to control ACK timing, then set `ackMode` to `MANUAL`.
```

modules/ROOT/pages/amqp/amqp-consume.adoc
```diff
@@ -5,44 +5,44 @@ endif::[]
...
+For more information about message ACKs, see xref:amqp/amqp-ack.adoc[How to handle Acknowledgment].
```
kbostrom:
```suggestion
For more information about message ACK operations, see xref:amqp/amqp-ack.adoc[How to Handle Acknowledgment].
```

modules/ROOT/pages/amqp/amqp-consume.adoc
```diff
@@ -5,44 +5,44 @@ endif::[]
...
 
 == Wait for a Message
 
-By default, the maximum wait time is configured to 10 seconds, producing an `AMQP:TIMEOUT` error if no message is available in that period.
-In order to configure a timeout that fits your use case better, customize the `maximumWait` and `maximumWaitUnit` parameters.
+By default, the maximum wait time is configured to 10 seconds. If no message is available during the specified wait time, an `AMQP:TIMEOUT` error is thrown. You can customize the wait time by configuring the `maximumWait` and `maximumWaitUnit` parameters.
 
-In order to wait indefinitely for a Message to arrive, the `maximumWait` value has to be set to `-1`. In this case, no `TIMEOUT` error will be thrown.
+To wait indefinitely for a Message to arrive, the `maximumWait` value must be set to `-1`. In this case, no `TIMEOUT` error will be thrown.
```
kbostrom:
```suggestion
To create an indefinite wait time for a message to arrive, set the `maximumWait` value to `-1`. In this case, no `TIMEOUT` error is thrown.
```

modules/ROOT/pages/amqp/amqp-consume.adoc
```diff
@@ -5,44 +5,44 @@ endif::[]
...
 
 == MIME Types and Encoding
 
-The AMQP Connector does its best to auto determine a message’s MIME type (`contentType`) based on the `contentType` property of the message. However, there are cases in which that best guess is not enough, and you need first-hand knowledge of the message’s content.
+AMQP Connector does its best to auto determine a message’s MIME type (`contentType`) based on the `contentType` property of the message. However, there are cases in which that best guess is not enough, and you need firsthand knowledge of the message’s content.
```
kbostrom:
```suggestion
AMQP Connector is designed to auto-determine a message’s MIME type (`contentType`) based on the `contentType` property of the message. However, there are cases in which it cannot make this determination, and you need firsthand knowledge of the message’s content.
```

modules/ROOT/pages/amqp/amqp-consume.adoc
```diff
@@ -5,44 +5,44 @@ endif::[]
...
 
-In such cases, you can force that content type to a particular value by using the `contentType` parameter.
+In such cases, you can force the content type to a particular value by using the `contentType` parameter.
 
-The same process works for encoding. By default, the connector will assume that the runtime’s default encoding matches the one in the message if no other information is provided. You can set this by using the `encoding` parameter.
+The same process works for encoding. By default, the connector assumes that the Mule's default encoding matches the one in the message if no other information is provided. You can set this by using the `encoding` parameter.
```
kbostrom:
```suggestion
The same process works for encoding. By default, the connector assumes that the default encoding of the Mule instance matches that of the message, if no other information is provided. You can set this by using the `encoding` parameter.
```

modules/ROOT/pages/amqp/amqp-consume.adoc
```diff
@@ -73,7 +73,7 @@ You can set the `createFallbackQueue` global config to prevent changes to the AM
 
 == Incoming Message Metadata
 
-As stated before, each message received will consist of two parts:
+As stated before, each message received consists of two parts:
```
kbostrom:
```suggestion
As stated earlier, each message received consists of two parts:
```

modules/ROOT/pages/amqp/amqp-listener.adoc
```diff
@@ -4,7 +4,7 @@ include::_attributes.adoc[]
 endif::[]
 :imagesdir: ../../assets/images/
 
-The Listener source in the AMQP connector provides the ability to consume messages as they arrive into an AMQP Queue.
+The Listener source in Anypoint Connector for AMQP (AMQP Connector) provides the ability to consume messages as they arrive into an AMQP Queue.
```
kbostrom:
```suggestion
The Listener source in Anypoint Connector for AMQP (AMQP Connector) provides the ability to consume messages as they arrive into an AMQP queue.
```

modules/ROOT/pages/amqp/amqp-publish-consume.adoc
```diff
@@ -5,7 +5,7 @@ endif::[]
 :imagesdir: ../../assets/images/
 
 
-Request and Reply is a very common pattern between applications, thus the AMQP Connector provides an operation specially designed for that, called `publish-consume`. The `publish-consume` operation allows you to publish a message to any exchange, and then wait for a reply on a different destination.
+Request and Reply is a very common pattern between applications, and AMQP Connector provides an operation specially designed for that, called `publish-consume`. The `publish-consume` operation allows you to publish a message to any exchange, and then wait for a reply on a different destination.
```
kbostrom: Wikipedia shows this as "request-reply" or "request-response."

kbostrom:
```suggestion
Request and Reply is a very common pattern between applications. AMQP Connector provides an operation specially designed for this, called `publish-consume`. The `publish-consume` operation enables you to publish a message to any exchange, and then wait for a reply on a different destination.
```

modules/ROOT/pages/amqp/amqp-publish.adoc
```diff
@@ -5,7 +5,7 @@ endif::[]
 :imagesdir: ../../assets/images/
 
 
-The Publish operation allows you to create a new AMQP message and send it to the specified exchange. You can then configure not only the content of the message, but also all the headers, the properties and the envelope in the AMQP message.
+The `publish` operation allows you to create a new AMQP message and send it to the specified exchange. You can then configure not only the content of the message, but also all the headers, the properties and the envelope in the AMQP message.
```
kbostrom:
```suggestion
The `publish` operation enables you to create a new AMQP message and send it to the specified exchange. You can then configure not only the content of the message, but also all the headers, the properties, and the envelope in the AMQP message.
```

kbostrom (approved):
Here you go!

PR #536, opened 2019-12-27
DOCS-6890

valkyrie69:
Bumped up subsections so they appear in the nav

### DIFF
--- a/modules/ROOT/pages/xml/xml-xpath.adoc
+++ b/modules/ROOT/pages/xml/xml-xpath.adoc
@@ -5,7 +5,7 @@ endif::[]
 
 The `<xml-module:xpath-extract>` operation supports the evaluation of XPath expressions.
 
-Because XPath expressions can match any number of individual elements, this operation returns a list of strings. If no element matches the expression, the operation returns an empty list.
+Because XPath expressions can match any number of individual elements, this operation returns a list of Strings. If no element matches the expression, the operation returns an empty list.
 
 XPath expressions are also namespace-aware, so this operation allows namespace mappings. These mappings are merged with any `namespaces` defined in the referenced `namespace-directory`, meaning that the evaluation combines both sets of namespace mappings.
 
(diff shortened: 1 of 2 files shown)


modules/ROOT/pages/xml/xml-xpath.adoc
```diff
@@ -139,7 +139,7 @@ There are cases in which the input XML document contains elements with different
 </soap:Envelope>
 ----
 
-The next example maps each `prefix` used in the XPath expressions to the corresponding namespace `uri`.
+The below example maps each `prefix` used in the XPath expressions to the corresponding namespace `uri`.
```
george-hoffer: Style says to not use above or below

modules/ROOT/pages/xml/xml-xpath.adoc
```diff
@@ -5,7 +5,7 @@ endif::[]
 
 The `<xml-module:xpath-extract>` operation supports the evaluation of XPath expressions.
 
-Because XPath expressions can match any number of individual elements, this operation returns a List of Strings. If no element matches the expression, the operation returns an empty list.
+Because XPath expressions can match any number of individual elements, this operation returns a list of strings. If no element matches the expression, the operation returns an empty list.
```
george-hoffer: Seems like it should be "list of Strings" -- to emphasize the use of the String class.

modules/ROOT/pages/xml/xml-xpath.adoc
```diff
@@ -221,10 +221,10 @@ Returning to the <<othello_lines, Othello lines>> example, imagine that you want
 ----
 
 * The first argument is the XPath expression.
-* The second one is the input document, in this case, the message payload.
-* The third one is the context properties. Because this case does not require any, the function passes an empty object (`{}`).
+* The second argument is the input document, which, in this case, is the message payload.
+* The third argument is the context properties, which in this case are not required, so the function passes an empty object (`{}`).
```
george-hoffer: these should be parallel. Use "which, in this case," or "which in this case" in both 224 and 225.

george-hoffer (approved):

PR #549, opened 2020-01-09
adding basic auth details for as2 connectors

bharathcp:


bharathcp:
The implementation has changed after architecture review. We have another PR for this change now - https://github.com/mulesoft/docs-connectors/pull/573/files

PR #558, opened 2020-01-13
Added aggregator listener limitation text.

dhirsch1:


modules/ROOT/pages/aggregator/aggregators-module-reference.adoc
```diff
@@ -265,6 +265,7 @@ Listeners are only able to reference aggregators that are inside a flow. If an a
 
 Once the aggregator that is referenced by the listener completes an aggregation, the listener is triggered with a list of all the elements. Though the aggregation listener can be used for any kind of aggregator, it is important for time-driven (Async) aggregations. Such aggregations are triggered asynchronously, so they do not execute an aggregator route and can only reach components in flows with an aggregator listener as the source.
 
+Using an aggregator listener creates a new flow. The context of the flow where the aggregator exists is not accessible from the new flow. Therefore, the new flow cannot access variables from the aggregrator's flow.
```
lucianoRM: The aggregator listener does not create a new flow. In any case, it should be: 
"Since the aggregator listener is a Source, it should be added as such in a different flow where the actual Aggregator is. The context of the flow where the aggregator...."

lucianoRM (changes requested):

modules/ROOT/pages/aggregator/aggregators-module-reference.adoc
```diff
@@ -265,7 +265,7 @@ Listeners are only able to reference aggregators that are inside a flow. If an a
 
 Once the aggregator that is referenced by the listener completes an aggregation, the listener is triggered with a list of all the elements. Though the aggregation listener can be used for any kind of aggregator, it is important for time-driven (Async) aggregations. Such aggregations are triggered asynchronously, so they do not execute an aggregator route and can only reach components in flows with an aggregator listener as the source.
 
-Using an aggregator listener creates a new flow. The context of the flow where the aggregator exists is not accessible from the new flow. Therefore, the new flow cannot access variables from the aggregrator's flow.
+Because the aggregator listener is a source, the context of the flow where the aggregator exists is not accessible unless you add the aggregator listener to that flow. Once you do this, the new flow can access variables from the aggregrator's flow.
```
lucianoRM: This is also not right. 
It should be something like:
Since the aggregator listener is a source, it will exist in a different flow from where the actual aggregator is.  Because of that, the context from the flow where the aggregator is is not accesible from the listener and, for example, variables are not propagated.

lucianoRM (changes requested):

lucianoRM (approved):

valkyrie69 (approved):

PR #586, opened 2020-01-23
Added entries for Ref Guides in Salesforce Connector landing pages.

dhirsch1:


dhirsch1:
Using another PR for this. Had trouble seeing the other updated Salesforce files when using this PR.

PR #630, opened 2020-02-06
Added note about Max redelivery

valkyrie69:


mbuchwald (commented):
Shouldn't it explain that this must be configured on the broker side?

mbuchwald (approved):

wwarner-salesforce (approved):

PR #643, opened 2020-02-14
Update object-store-connector-reference.adoc

praveenbandaru:
The description of the Key under Contains was incorrect.


hannanelson:
Verified correction with @sajajdo. See #665.

PR #644, opened 2020-02-15
Update object-store-connector.adoc

GreatGangi:
Incorrect: Note that the Mule app is deployed into CloudHub workers through Anypoint Platform Runtime Manager, the contents of the Mule app's default Object Store are visible in Runtime Manager, within the Mule app's Application Data tab.

Correct: Note that the Mule app is deployed into CloudHub workers through Anypoint Platform Runtime Manager, the contents of the Mule app's default Object Store are invisible in Runtime Manager, within the Mule app's Application Data tab.


hannanelson:
Verified correction with @estebanwasinger. See #665.

PR #655, opened 2020-02-21
Update mongodb-connector-reference.adoc

santiagoGQ:


dhirsch1:
PR is not needed. I put the information in the Release Notes since there is no Reference Guide for the version that this PR applies to.

PR #678, opened 2020-03-11
DOCS-7607 WIP: Authentication Topic for Anypoint Connector Introduction

valkyrie69:


modules/ROOT/pages/introduction/anypoint-connector-authentication.adoc
```diff
@@ -0,0 +1,23 @@
+= Introduction to Anypoint Connectors
```
dhirsch1, line 1: General comments: 

Do we want to distinguish between OAuth 1.0 and OAuth 2.0 or at least mention them? I think some connectors still support OAuth 1.0. Also, some connectors call Basic authentication User and Password authentication. I will try to verify this.

I wonder if this information should go into the Configuration Overview topic that I just sent to Kim for editing. I mention authentication there when I talk about the global element.

valkyrie69:
After discussing with Diane, we have decided to not have this as a high level intro topic.

PR #724, opened 2020-04-23
Added authentication information to the Studio Topic for Cassandra

dhirsch1:


valkyrie69 (approved):
Per our updates to the template, we are changing the title names to: 

Use Flow Designer to Configure a Connector
and
Use Studio to Configure a Connector

If you make those updates, you will have to update the title in the nav doc too. We can probably write a script to update all of the titles at once though, so it's up to you.

PR #778, opened 2020-06-01
Update x12-edi-schema-language-reference.adoc

mcherem:


modules/ROOT/pages/x12-edi/x12-edi-schema-language-reference.adoc
```diff
@@ -84,7 +84,9 @@ The values are:
 * C for Conditional
 * M for Mandatory
 * O for Optional
-* U for Unused
+* U for Unused, the item should not appear at all in data
+* I for Ignored, it may be present in data (in the case of the HL segments it must be present) but just doesn't show up for the user in metadata
```
dhirsch1: Global:

I think you need a comma after "data."
My opinion (that you can ignore) is that "the data" is better than "data." I would also remove "just."

dhirsch1 (approved):

PR #796, opened 2020-06-16
DOCS-7313 x12 connector template

valkyrie69:


valkyrie69:
This is very outdated. This needs to be updated with the current template.

PR #879, opened 2020-07-14
Update sap-connector-reference.adoc

valean27:


amitfernandes:
Approved the change to call out that s/4HANA on-premises is also supported.

hannanelson (approved):

PR #882, opened 2020-07-15
Updating Oracle EBS (compatibility with Java 8)

mauro-dentoni:
The connector should be JDK 11 compatible, by End-July


kbostrom (approved):
Approved

PR #892, opened 2020-07-20
Added mule4 example for upsert parent child relationship

alexandruatomei9:


salesforce/10.2/modules/ROOT/pages/salesforce-connector-examples.adoc
```diff
@@ -267,110 +267,81 @@ A Salesforce object can have standard or custom relationships between objects.
 
 The relationships between the objects are usually one-to-many parent child relationships, but can be any link between two objects residing in Salesforce.
 
-Creating or altering objects with relationships is challenging. This section shows how to perform an upsert for an object using Salesforce Connector.
+Creating or altering objects with relationships is challenging. This example shows how to create the object relationship structure in Salesforce to perform an upsert for an object using Salesforce Connector.
```
stefangretcan, line 270: Since we are stating that this example shows how to create the relationship structure in Salesforce, we cannot pretend on line #272 that Plance_c should already exist. We either just say that the objects already exist and what is the relationship between them or we need to show the user how to create them from scratch.

salesforce/10.2/modules/ROOT/pages/salesforce-connector-examples.adoc
```diff
@@ -267,110 +267,81 @@ A Salesforce object can have standard or custom relationships between objects.
...
 
-Create a structure in Salesforce for this relationship.
+This example uses two custom Salesforce object types: `Component__c` and `Plane__c`. The `Plane__c` entity must already exist in your environment.
```
stefangretcan, line 272: I think the custom object names are not very descriptive. We can find out a better scenario, maybe the same that is described here: https://trailhead.salesforce.com/en/content/learn/modules/data_modeling/object_relationships, the Propery --> offer one.

dhirsch1 (approved):

PR #911, opened 2020-07-24
fix remaining broken xrefs

mojavelinux:


mojavelinux:
@sleslie23 last two, hopefully

mojavelinux:
Even though these changes match the most recent commit, they are still relevant since I think they are the more correct links.

sleslie23:
This was fixed by another PR.

PR #980, opened 2020-08-19
Update email-send.adoc

Arges86:
Added the `CDATA`, otherwise the xml parser treats the 'H1' tag as child element


kevintroller:
Thank you for reporting this change. Closing as worked in https://github.com/mulesoft/docs-connectors/pull/987

PR #1037, opened 2020-09-14
DOCS-8925 TTL in OSv2 Connector Reference

hannanelson:


object-store/1.1/modules/ROOT/pages/object-store-connector-reference.adoc
```diff
@@ -345,6 +346,82 @@ If `Expiration Interval` is set to a value lower than or equal to zero, then the
...
++
+** Mule version 4.2.x, 4.3.0, and later: Rolling TTL
+** Mule versions earlier than 4.2.1: Static TTL of 30 days by default
+
+* Less than or equal to `0`
++
+TTL set to the maximum value (2592000 seconds/30 days).
+* Positive number
++
+TTL set to the specified value
```
podviaznikov: maybe add something  like "but not more than max TTL (2592000 seconds/30 days)"

object-store/1.1/modules/ROOT/pages/object-store-connector-reference.adoc
```diff
@@ -345,6 +346,82 @@ If `Expiration Interval` is set to a value lower than or equal to zero, then the
...
+
+// ** Mule version 4.2.x, 4.3.0, and later: TTL set to the specified value and creates a rolling TTL.
```
podviaznikov: I realized that 4.2.x is not correct. Because there is 4.2.0 - that we never touched.

So should be >=4.2.1, 4.3.x. Not sure what is the best to write that

hannanelson: 4.2.1 and later

object-store/1.1/modules/ROOT/pages/object-store-connector-reference.adoc
hannanelson: added (between 1 and 2592000 seconds)

object-store/1.1/modules/ROOT/pages/object-store-connector-reference.adoc
```diff
@@ -345,6 +346,79 @@ If `Expiration Interval` is set to a value lower than or equal to zero, then the
...
+If the field contains: 
+
+* No value
++
+** Mule versions 4.2.1 and later: Rolling TTL
+** Mule versions earlier than 4.2.1: Static TTL of 30 days by default
+
+* Less than or equal to `0`
++
+TTL set to the maximum value (2592000 seconds/30 days).
```
schinthakindims: Also, specify that this will not be rolling ttl in this case.

schinthakindims (changes requested):

object-store/1.1/modules/ROOT/pages/object-store-connector-reference.adoc
hannanelson: added static

schinthakindims (changes requested):
LGTM

object-store/1.1/modules/ROOT/pages/object-store-connector-reference.adoc
```diff
@@ -345,6 +346,79 @@ If `Expiration Interval` is set to a value lower than or equal to zero, then the
...
+=== Object Store (Object Store v2)
+
+
+NOTE: These settings apply only to xref:object-store::index.adoc[Object Store v2].
+
+
+[%header%autowidth.spread]
+|===
+| Field | Type | Description | Default Value | Required
+| Persistent a| Boolean | Whether the store is persistent or transient. | true |
```
kbostrom, line 361: I wonder about that Required column. 
It looks like it is rarely used.
Would it be a good idea to simply replace it with specifying whether a field is required when naming the field?
Or, perhaps we should use "Yes" and "No," so that a blank cell does not appear? I always wonder if audiences question a blank cell, which is why most depts. I have worked in use "n.a." to reassure readers that something isn't just missing....

object-store/1.1/modules/ROOT/pages/object-store-connector-reference.adoc
```diff
@@ -345,6 +346,79 @@ If `Expiration Interval` is set to a value lower than or equal to zero, then the
...
+* Less than or equal to `0`
++
+TTL is static and set to the maximum value (2592000 seconds/30 days).
+* Positive number
++
+TTL set to the specified value (between 1 and 2592000 seconds)
+
+*Include link to TTL in FAQ and/or TTL task*
+| 2592000 seconds (30 days)  |
+| Entry Ttl Unit a| Enumeration, one of:
```
kbostrom: Ttl lettercasing OK here?

object-store/1.1/modules/ROOT/pages/object-store-connector-reference.adoc
```diff
@@ -345,6 +346,79 @@ If `Expiration Interval` is set to a value lower than or equal to zero, then the
...
+
+** NANOSECONDS
+** MICROSECONDS
+** MILLISECONDS
+** SECONDS
+** MINUTES
+** HOURS
+** DAYS a| The unit for the entry's time to live (`entryTtl`) | SECONDS |
```
kbostrom: Should this end with a period; is it a peer with the other fragments that end in periods above?

object-store/1.1/modules/ROOT/pages/object-store-connector-reference.adoc
```diff
@@ -345,6 +346,79 @@ If `Expiration Interval` is set to a value lower than or equal to zero, then the
...
+** NANOSECONDS
+** MICROSECONDS
+** MILLISECONDS
+** SECONDS
+** MINUTES
+** HOURS
+** DAYS a| The expiration interval (`expirationInterval`). | MINUTES |
+| Config Ref a| <<ObjectStoreConnector>> a| A reference to an `os:config` element used to declare this Object Store.
+
+If not provided, the Mule runtime's default ObjectStoreManager is used. Setting this parameter is only necessary when you want to use a non-default Object Store implementation. For example, if you want to have a store which is backed by Redis or JDBC. |  |
```
kbostrom:
```suggestion
If not provided, the Mule runtime's default ObjectStoreManager is used. Setting this parameter is necessary only when you want to use a nondefault Object Store implementation: for example, if you want to have a store that is backed by Redis or JDBC. |  |
```

object-store/1.1/modules/ROOT/pages/object-store-connector-reference.adoc
```diff
@@ -345,6 +346,79 @@ If `Expiration Interval` is set to a value lower than or equal to zero, then the
...
+| Expiration Interval Unit a| Enumeration, one of:
+
+** NANOSECONDS
+** MICROSECONDS
+** MILLISECONDS
+** SECONDS
+** MINUTES
+** HOURS
+** DAYS a| The expiration interval (`expirationInterval`). | MINUTES |
+| Config Ref a| <<ObjectStoreConnector>> a| A reference to an `os:config` element used to declare this Object Store.
```
kbostrom: GLOBALLY be sure that we are making the distinction between object store and Object Store clear and consistent.

kbostrom (dismissed):
Edited per scope of request

object-store/1.1/modules/ROOT/pages/object-store-connector-reference.adoc
hannanelson: It's a mishmash of the field/parameter names. I'll use field names, which are also questionably capitalized.

schinthakindims (approved):

raju-sfdc (approved):

kbostrom (approved):
Reapproval scan of terminology presentation changes

PR #1045, opened 2020-09-21
Add note regarding payload tranformations in cryptographic operations

fermujica:


wss/1.0/modules/ROOT/partials/transformation-note.adoc
```diff
@@ -0,0 +1,4 @@
+[NOTE]
+--
+Transform operations don't take into account the cryptographic signatures and can end up modifying them.
```
rddheath: If this note applies to transform operations in general, you can remove "the" in front of "cryptographic signatures".

rddheath: What do you mean by "take into account"?

rddheath: "end up"?
Perhaps something like this? "Transform operations ... and might change them."
But now it seems as though the reader needs some explanation of the sorts of changes that transform operations might make to cryptographic signatures.

rddheath (approved):

rddheath (approved):

wss/1.0/modules/ROOT/partials/transformation-note.adoc
adobni: Any transformation to the content can break the cryptographic signatures. It is not obvious to users, but DataWeave doesn't just output the input as-is. It will parse the input, convert to an internal format, and generate a new output. For other use cases it might be indistinguishably from the input for XML processing purposes. Just doing #[payload] for example emits a new output, that can change the formatting and break the signature validations.

adobni: Ignores them as any other XML values. It will not try to preserve formatting.

PR #1114, opened 2020-10-27
Added detail from mule-runtime

hannanelson:


prafulrana (dismissed):

schinthakindims (commented):
If we are with duplicating this information (from here https://docs.mulesoft.com/mule-runtime/4.3/redelivery-policy), my concern is with maintaining this page correctly and keeping it up to date. Should we just refer the whole section from the mule page?

schinthakindims (approved):
LGTM

sarathecoubian (approved):

PR #1125, opened 2020-11-03
Update salesforce-connector-reference.adoc

lopezmas:
Parameter queryParams is not documented for Apex Rest calls using the Salesforce Connector. This is critical information to be able to call SF Apex Rest endpoints that expects URL parameters.


alexandruatomei9 (approved):

valkyrie69:
This update will be in PR #1152

PR #1148, opened 2020-11-17
DOCS-9479-db-troubleshooting-kt

kevintroller:
DB 1.9 database-troubleshooting.adoc editorial


db/1.9/modules/ROOT/edits/database-connector-troubleshooting.adoc
```diff
@@ -0,0 +1,47 @@
+= Database Connector Troubleshooting - Mule 4
+
+To troubleshoot Anypoint Connector for Database (Database Connector), review the following causes and possible solutions.
```
kbostrom: Instead of "causes" I would use "issues."
Users aren't as interested in what causes an issue as what they are seeing as an error message or an unexpected result: the "issue" they are experiencing.

kbostrom: ...and if there is just one...:
```suggestion
To troubleshoot Anypoint Connector for Database (Database Connector), review the following issue and possible solution.
```
In fact, if there is just one, you could promote the content titles:
```suggestion
= Troubleshoot a Flow That Hangs
Using a pooling profile....
One solutions is....
```
...but I suspect you want to allow for other Troubleshooting topics down the road?

db/1.9/modules/ROOT/edits/database-connector-troubleshooting.adoc
```diff
@@ -0,0 +1,47 @@
...
+
+== Pooling Profiles Connections Making the Flow Hang
+Using a pooling profile that provides fewer connections than what the flow uses can make the entire flow hang.
```
kbostrom:
```suggestion
Using a pooling profile that provides fewer connections than the flow requires can make the entire flow hang.
```
"can" or "does"?

db/1.9/modules/ROOT/edits/database-connector-troubleshooting.adoc
```diff
@@ -0,0 +1,47 @@
...
+
+=== Solution
+One solution is to increase the size of the pool. However, sometimes that's not possible, so another solution is to use the VM Connector to shorten the life of the connection to be the same as the life of a VM message. This way, every call to the *Stored procedure* operation is executed, its result is consumed, and the connection associated with the operation is closed and returned to the pool.
```
kbostrom:
```suggestion
One solution is to increase the size of the pool. However, sometimes that's not possible, so another solution is to use Anypoint Connector for VM to shorten the life of the connection to match the life of a VM message. This way, every call to the *Stored procedure* operation is executed, its result is consumed, and the connection associated with the operation is closed and returned to the pool.
```

db/1.9/modules/ROOT/edits/database-connector-troubleshooting.adoc
```diff
@@ -0,0 +1,47 @@
...
+
+The following example shows how to use pooling profiles with the *Stored procedure* operation and VM Connector. When the `foreach` element reaches the eleventh iteration the *Stored procedure* operation won't hang since the pool is able to provide a connection:
```
kbostrom:
```suggestion
The following example shows how to use pooling profiles with the *Stored procedure* operation and VM Connector. In this example, when the `foreach` element reaches the eleventh iteration, the *Stored procedure* operation doesn't hang, because the pool is able to provide a connection:
```

kbostrom (dismissed):
CE as requested

dhirsch1 (approved):

PR #1164, opened 2020-11-26
Update sap-s4hana-cloud-connector-reference.adoc

xidex:
- Updated javadoc of service discovery related parameters.


mcherem (dismissed):
All looks good - just one quick question. For the:
Discovery Service Enabled, I notice you removed the disabled explanation. I'm assuming you felt it was redundant and unnecessary?

mcherem (approved):
Peer review

PR #1191, opened 2020-12-17
DOCS-9489-http-load-example-k

kevintroller:
Updated with new template, Studio steps, screenshots and XML.


http/1.5/modules/ROOT/pages/http-load-static-res-task.adoc
```diff
@@ -1,29 +1,72 @@
...
 :keywords: anypoint, connectors, transports
 :page-aliases: connectors::http/http-load-static-res-task.adoc
 
-. Create a new project.
-. From the Mule Palette, drag an HTTP Listener operation, to a new flow.
-. Configure the Listener:
-* Click + and set the default host, port and base path. For example, host = 0.0.0.0, port = 8081, and no base path.
-* In General, set Path to the URL you plan to use to start the app from the browser. For example, set Path to */trigger*.
-* Accept the other default Listener settings, and save the configuration.
+The following example illustrates how to configure the Anypoint Connector for HTTP (HTTP Connector) *Listener* source and *Load static resource* operation to serve up static content for use with HTTP, using the request path to lookup the resource.
```
mcherem: Question - for some connectors, I see language such as: In this example... Would it make sense to reword this here?
In this example, you configure the Anypoint Connector for HTTP (HTTP Connector) *Listener* source and *Load static resource* operation to serve up static content for use with HTTP by using the request path to lookup the resource.
This may be the way Kim wants it stated, so if that's the case, please ignore.

http/1.5/modules/ROOT/pages/http-load-static-res-task.adoc
```diff
@@ -1,29 +1,72 @@
...
+To accomplish this example, you must create the Mule app, configure an HTTP global element, run, and test the app with curl commands.
```
mcherem, line 6: Question here - is this an overview of what the user needs to do before completing the example? I'm asking because this didn't seem like an overview of the steps the user needed to complete for the example. That is, does the user need to do these first? If so, then do they need a reference to the other file for how to do these items? Or, am I missing something here? Maybe that's how Kim would like to see this info presented. In that case, please ignore.

mcherem (approved):
Couple of things, but otherwise, great!

http/1.5/modules/ROOT/pages/http-load-static-res-task.adoc
```diff
@@ -1,29 +1,72 @@
...
+The following screenshot shows the Anypoint Studio app flow for this example:
+
+.Load a static resource flow
+image::http-loadresource-flow.png[Load a static resource flow]
+
+== Create the Mule Application
+
+To create the Mule flow:
+
+. In the *Mule Palette* view, select the HTTP *Listener* source and drag it on to the canvas. +
```
mcherem: For an easy way to remember when to use 'on to' or 'onto'. The main difference between the two is that onto is about movement from one place to another. The word 'on' is not.

mcherem (approved):

http/1.5/modules/ROOT/pages/http-load-static-res-task.adoc
mcherem: 'the' before the 'load static...'
Also, globally check to make sure there is correct article placement (such as 'the' or 'a')

http/1.5/modules/ROOT/pages/http-load-static-res-task.adoc
```diff
@@ -1,29 +1,72 @@
-= To Load a Static Resource
+= Load a Static Resource Example - Mule 4
```
mcherem, line 1: Some global items to check for:

- Use of 'in' instead of 'on' as with references to pathnames etc.
- Remember to insert articles ('the' and 'a') as appropriate.
- Use 'that' instead of 'which' in most cases.
- Watch your comma placement.
- If a name is not a product name, then it does not require init caps

mcherem (dismissed):

mcherem (approved):

PR #1203, opened 2020-12-30
Added new configurations in writer tab for 2.1.5 version

sms-mgrunblatt:


mcherem:
There is a new PR for this work: DOCS-9797-EDIFACT-2.2 This is because a new version of the docs has been created for this feature release: 2.2 and these docs are now in the new template structure. Thus, the location of the information to be changed has moved.

PR #1208, opened 2021-01-05
Minor addition to salesforce doc

alexandruatomei9:


valkyrie69 (approved):

bauerchristina:
In my opinion it should not say " Maximum value for the event size is 2147483647", but " The maximum value for the event **queue** size is 2147483647".

PR #1216, opened 2021-01-13
DOCS-9935 Veeva Vault 1.3 

valkyrie69:


veevavault/1.3/antora.yml
```diff
@@ -0,0 +1,16 @@
+name: veevavault-connector
+version: '1.2'
```
dhirsch1, line 2: Should be 1.3 here and below

dhirsch1 (approved):

PR #1237, opened 2021-01-23
Update kafka-connector-studio.adoc

eduardosponda:
The documentation forgets about the offset for the Seek operation and what is this Input Source section about? The Seek operation also doesn't have a screenshot of the connector config screen.


valkyrie69:
This issue was addressed in PR #1284 Thank you for pointing it out.

PR #1238, opened 2021-01-25
Doc edit for the Slack Connector docs

dhirsch1:


dhirsch1:
Could not get the Examples file to be green so it can be edited. Will try again with another PR.

PR #1245, opened 2021-01-28
DOCS-10040 database-select-example-kt

kevintroller:
- Rewrote Database Connector -  Select operation documentation
- Added Studio steps and screenshots
- Reorganized content


db/1.9/modules/ROOT/pages/database-connector-query-single.adoc
```diff
@@ -5,7 +5,7 @@ The *Query single* operation enables you to execute an SQL query that returns a
 You must be careful, however, because if your row is too large, you could run out of memory.
 Also, if you use connection pooling you do not have to worry about your application running out of connections, because this operation returns the connection to the pool immediately after the query is executed.
 
-Although you invoke the *Query single* operation in the same way as you invoke the *Select* operation, although you always get a single result, regardless of the number of records returned by the actual SQL query.
+Although you invoke the *Query single* operation in the same way as you invoke the *Select* operation, you always get a single result, regardless of the number of records returned by the actual SQL query.
```
kbostrom: Since I am missing the context here, let me just check: are we using "Although" because the *Select* operation will give more than one result? Maybe...:
```suggestion
Although you invoke the *Query single* operation in the same way as you invoke the *Select* operation, the result is not the same; whereas *Select* might return more than one result, *Query single* always returns a single result, regardless of the number of records returned by the actual SQL query.
```

kbostrom: This might be overkill, since "query single" implies either one query or one result.
Another option is...:
```suggestion
You invoke the *Query single* operation in the same way as you invoke the *Select* operation; however, you always get a single result using *Query single*, regardless of the number of records returned by the actual SQL query.
```

db/1.9/modules/ROOT/pages/database-connector-select.adoc
```diff
@@ -0,0 +1,184 @@
...
+
+== Configure the Select Operation in Studio
+
+To add and configure the Select operation in Studio, follow these steps:
+
+. In the *Mule Palette* view, search for `database` and select the *Select* operation.
+. Drag the *Select* operation onto the Studio canvas.
+. In the *General* tab of the operation configuration screen, click the plus sign (*+*) next to the *Connector configuration* field to access the global element configuration fields.
+. Specify the database connection information and click *OK*.
+. Set the *SQL Query Text* field to the SQL query to execute, for example `SELECT * FROM PLANET WHERE name = :name`.
```
kbostrom:
```suggestion
. Set the *SQL Query Text* field to the SQL query to execute, for example, `SELECT * FROM PLANET WHERE name = :name`.
```

db/1.9/modules/ROOT/pages/database-connector-select.adoc
```diff
@@ -0,0 +1,184 @@
+= Query a Database Examples - Mule 4
+:page-aliases: connectors::db/db-connector-query.adoc
+
+Anypoint Connector for Database (Database Connector) Select operation selects data from a database. These examples help you configure the operation in Studio and XML.
+
```
kbostrom, line 5: "These examples help you configure the operation in Studio and XML" led me to believe that, given the following topic, Configure the Select Operation in Studio, that I would then see Configure the Select Operation in XML...which I don't see here?

Be sure to word your intro text at Level 1 so that the audience expect sprecisely what is contained at subordinate levels.

db/1.9/modules/ROOT/pages/database-connector-select.adoc
```diff
@@ -0,0 +1,184 @@
...
+In the XML editor, the `<db:select>` configuration looks like this:
+[source,xml,linenums]
+----
+<db:select config-ref="dbConfig">
+  <db:sql>SELECT * FROM PLANET WHERE name = :name</db:sql>
+</db:select>
+----
+
+== Configure the Input Parameters Field
```
kbostrom: This is a peer topic to Configure the Select Operation in Studio?

db/1.9/modules/ROOT/pages/database-connector-select.adoc
```diff
@@ -0,0 +1,184 @@
...
+
+To protect database queries, configure the *Input Parameters* field in the Select operation by adding variable values to the SQL statement you execute in the database. The primary goal of the Select operation is to supply a SQL query and use DataWeave for the parameters.
+
+The parameters you provide are a map in which keys are the name of an input parameter to be set on the JDBC prepared statement. Each parameter should be referenced in the SQL text using a colon prefix, for example `where id = :myParamName`. The map’s values contains the actual assignation for each parameter.
```
kbostrom: "The parameters are a map"? Or they create a map?

kbostrom: passive voice: "Each parameter should be referenced in the SQL text using a colon prefix...."
I've seen this earlier, and I missed commenting about it; can we reword to be active, or is it misleading to use something like "Reference each parameter in the SQL text by using a colon prefix"?

kbostrom: number agreement: "values contains" > "values contain" or "value contains"

db/1.9/modules/ROOT/pages/database-connector-select.adoc
```diff
@@ -0,0 +1,184 @@
...
+
+The advantages of using input parameters to configure the `WHERE` clause in a SQL `SELECT` statement are that it makes the query immune to SQL injection attacks and enables optimizations that are not possible otherwise improving the application performance.
```
kbostrom: comma
```suggestion
The advantages of using input parameters to configure the `WHERE` clause in a SQL `SELECT` statement are that it makes the query immune to SQL injection attacks and enables optimizations that are not possible otherwise, improving the application performance.
```

kbostrom: Shorter, more active version:
```suggestion
Use input parameters to configure the `WHERE` clause in a SQL `SELECT` statement to make the query immune to SQL injection attacks and to enable optimizations that are not possible otherwise, improving the application performance.
```

db/1.9/modules/ROOT/pages/database-connector-select.adoc
```diff
@@ -0,0 +1,184 @@
...
+
+For security reasons, do not directly write `<db:sql>SELECT * FROM PLANET WHERE name = #[payload] </db:sql>`.
+
+DataSense is available for the operation's input and output. Database Connector analyzes the query and automatically calculates the structure of the query's output by analyzing the projection section of the SQL statement. At the same time, by comparing the conditions in the `WHERE` clause to the table structure, it generates DataSense input to help you build the DataWeave script that generates the input parameters.
```
kbostrom: DataSense > DataSense Explorer? DataSense Query Language?

kbostrom: What does it mean to say that DataSense "is available for input and output"? available to use to create input and output? to receive input and output? Or some other action?

db/1.9/modules/ROOT/pages/database-connector-select.adoc
```diff
@@ -0,0 +1,184 @@
...
+
+In the following example, you supply input parameters as key-value pairs embedding a DataWeave script. The keys use the colon character (`:`) to reference a parameter value by name:
```
kbostrom: What is "embedding"? Do we mean that the pairs are embedded in the script, or that the script is embedded in the pairs?

db/1.9/modules/ROOT/pages/database-connector-select.adoc
```diff
@@ -0,0 +1,184 @@
...
+
+. In your Studio flow, select the *Select* operation.
+. In the operation configuration screen, set the *SQL Query Text* field to `SELECT * FROM PLANET WHERE name = :name`.
+. Set the *Input Parameters* field to `{'name' : payload}`.
+
+The following screenshot shows the *Input Parameters*  configuration:
```
kbostrom: extra space
```suggestion
The following screenshot shows the *Input Parameters* configuration:
```

db/1.9/modules/ROOT/pages/database-connector-select.adoc
kbostrom: Is this a more appropriate place to mention the text of the first Note below, if some users have to escape the colon?

db/1.9/modules/ROOT/pages/database-connector-select.adoc
```diff
@@ -0,0 +1,184 @@
...
+<flow name="DatabaseFlow">
+  <db:select config-ref="dbConfig">
+    <db:sql>SELECT * FROM PLANET WHERE name = :name</db:sql>
+    <db:input-parameters>
+      #[{'name' : payload}]
+    </db:input-parameters>
+  </db:select>
+</flow>
+----
+
```
kbostrom, line 65: Instead of NOTEs, I suggest that the following be converted to text about version-specific considerations: perhaps an intro statement and then a two-item definition list? Avoid using boldface to create narrative headings.

kbostrom: I also wonder if these NOTE texts should appear before the task, kind of like prerequisites or requirements?

kbostrom: NOTES copyedited only; consider alternatives above, or other ways to insert this info into main narrative instead of tacking it onto the end of the task.

db/1.9/modules/ROOT/pages/database-connector-select.adoc
```diff
@@ -0,0 +1,184 @@
...
+[NOTE]
+====
+*Since version 1.4.0 (escaping colons)*:
```
kbostrom: Escaping Colons in SQL Queries in <product?> 1.4.0 and Later

db/1.9/modules/ROOT/pages/database-connector-select.adoc
```diff
@@ -0,0 +1,184 @@
...
+If you need to use the colon character (`:`) in your SQL query, you can escape it
```
kbostrom: "...you can escape it" or "you must"? or simply "escape"?
```suggestion
If you use the colon character (`:`) in your SQL query, escape it
```

db/1.9/modules/ROOT/pages/database-connector-select.adoc
```diff
@@ -0,0 +1,184 @@
...
+by putting a backslash before it. This is useful when using PostgreSQL type
+casting, which requires two colons before the type you are casting to, for example:
+
+`<db:sql>SELECT price\:\:float8 FROM PRODUCT</db:sql>`
+====
+
+[NOTE]
+====
+*Since version 1.8.0 (SQL Casting for PostgreSQL and Snowflake)*:
```
kbostrom: SQL Casting for PostgreSQL and Snowflake in <product> 1.8.0

db/1.9/modules/ROOT/pages/database-connector-select.adoc
```diff
@@ -0,0 +1,184 @@
...
+Database Connector now accepts SQL Casting PostgreSQL and Snowflake syntax's, through the double colon (::) expression without the need to escape each colon (this feature doesn't affect the escaping colons behaviour). For example:
```
kbostrom:
```suggestion
Database Connector 1.8.0 and later accepts SQL Casting PostgreSQL and Snowflake syntaxes through the double colon (::) expression without escaping each colon and without affecting the behavior of escaped colons, for example:
```
This is a little unclear to me. Does this imply that the issue in the previous note is no longer an issue in 1.8.0 and later? And we say that there is no need to escape each colon, but that this does not affect...colons that are escaped?

db/1.9/modules/ROOT/pages/database-connector-select.adoc
```diff
@@ -0,0 +1,184 @@
...
+
+`<db:sql>SELECT MAX(modified_date)::DATE FROM sales</db:sql>`
+====
+
+
+== Configure Dynamic Queries
```
kbostrom: ditto re. peer title to first Level-2 subtopic? GLOBAL for all ==s that follow.

db/1.9/modules/ROOT/pages/database-connector-select.adoc
```diff
@@ -0,0 +1,184 @@
...
+
+When you need to parameterize not only the `WHERE` clause, but also parts of the query itself, for example queries that need to compare tables that depend on a condition, or complex queries for which the project table columns need to vary, you can configure dynamic queries.
```
kbostrom: This is a LOT of modification for one sentence. Consider making it more than one. Or use parens for some of the modification, to prevent confusion around all the commas:
```suggestion
When you need to parameterize not only the `WHERE` clause but also parts of the query itself (for example, queries that compare tables that depend on a condition, or complex queries for which the project table columns vary), you can configure dynamic queries.
```

db/1.9/modules/ROOT/pages/database-connector-select.adoc
```diff
@@ -0,0 +1,184 @@
...
+
+In the following example, you configure a dynamic query by using a full expression with a string in which the table depends on a variable `$(vars.table)`. Although, some of the query text is dynamic (`"SELECT * FROM $(vars.table)`), the `WHERE` clause still defines the `WHERE` condition using input parameters, in this case, `WHERE name = :name`.
```
kbostrom: Never use a comma after "although":
```suggestion
In the following example, you configure a dynamic query by using a full expression with a string in which the table depends on a variable `$(vars.table)`. Although some of the query text is dynamic (`"SELECT * FROM $(vars.table)`), the `WHERE` clause still defines the `WHERE` condition using input parameters: in this case, `WHERE name = :name`.
```

db/1.9/modules/ROOT/pages/database-connector-select.adoc
```diff
@@ -0,0 +1,184 @@
...
+<set-variable variableName="table" value="PLANET"/>
+<db:select config-ref="dbConfig">
+    <db:sql>#["SELECT * FROM $(vars.table) WHERE name = :name"]</db:sql>
+    <db:input-parameters>
+        #[{'name' : payload}]
+    </db:input-parameters>
+</db:select>
+----
+
+Input parameters can be applied only to parameters in a `WHERE` clause. To modify any other part of the query, use the DataWeave’s interpolation operator.
```
kbostrom: passive, "the DataWeave's" > "the DataWeave" or "DataWeave's":
```suggestion
You can apply input parameters only to parameters in a `WHERE` clause. To modify any other part of the query, use the DataWeave interpolation operator.
```

db/1.9/modules/ROOT/pages/database-connector-select.adoc
```diff
@@ -0,0 +1,184 @@
...
+
+
+== Stream Large Results
+
+Use streaming with queries that return many records, such as in integration use cases. In Mule 4, streaming is transparent and always enabled.
```
kbostrom, line 114: Can we preceded this with the reason? For example, "For better performance, use...."?

db/1.9/modules/ROOT/pages/database-connector-select.adoc
```diff
@@ -0,0 +1,184 @@
...
+
+For example, if you submit a query that returns 10K rows by attempting to fetch all those rows at once results in both performance degradation, due to the big pull from the network, and the risk of running out of memory, because all the information must be loaded into RAM.
```
kbostrom, line 116: Simplify sentence and I am assuming that your audience understands what causes the degradation and memory issues in this case:
```suggestion
For example, if you submit a query that returns 10K rows, attempting to fetch all those rows at once results in both performance degradation and the risk of running out of memory.
```

db/1.9/modules/ROOT/pages/database-connector-select.adoc
```diff
@@ -0,0 +1,184 @@
...
+
+With streaming, Database Connector fetches and processes only part of the query at one time, reducing the load on the network and memory.  This means that the connector does not fetch the 10K rows at once; instead, it fetches a smaller chunk, and once that chunk is consumed, it fetches the rest.
```
kbostrom, line 118: Is this para really necessary? Even to a neophyte like myself, the previous paragraph implies this.

db/1.9/modules/ROOT/pages/database-connector-select.adoc
```diff
@@ -0,0 +1,184 @@
...
+
+You can also use the new repeatable streams mechanism where DataWeave and other components process the same stream many times, even in parallel.
```
kbostrom: Maybe avoid "new," because at some point, this is not going to be new?
```suggestion
You can also use the repeatable streams mechanism, which enables DataWeave and other components to process the same stream many times, even in parallel.
```

db/1.9/modules/ROOT/pages/database-connector-select.adoc
```diff
@@ -0,0 +1,184 @@
...
+
+== Configure the Fetch Size and Max Rows Fields
+
+Because Mule runtime engine (Mule) enables Database Connector to manage streaming that does not mean that it's a good idea to move large chunks of data from the database to Mule. Even with streaming, a simple SQL query can return many rows, each one containing a lot of information. To limit the results, you can configure the *Fetch size* and *Max rows* fields.
```
kbostrom:
```suggestion
Although Mule runtime engine (Mule) enables Database Connector to manage streaming, it's not always a good idea to move large chunks of data from the database to Mule. Even with streaming, a simple SQL query can return many rows, each one containing a lot of information. To limit the results, configure the *Fetch size* and *Max rows* fields.
```

db/1.9/modules/ROOT/pages/database-connector-select.adoc
```diff
@@ -0,0 +1,184 @@
...
+
+In the following example, you configure these fields for the Select operation. The syntax instructs Database Connector to fetch no more than 1000 rows (*Max rows* value), no more than 200 rows at a time (*Fetch size* value), significantly reducing network and memory load. The *Fetch size* value is enforced differently by different JDBC driver providers and often defaults to `10`.
```
kbostrom:
```suggestion
In the following example, you configure these fields for the Select operation. The syntax instructs Database Connector to fetch no more than 1000 rows (*Max rows* value) and no more than 200 rows simultaneously (*Fetch size* value), significantly reducing network and memory load. The *Fetch size* value is enforced differently by different JDBC driver providers and often defaults to `10`.
```

db/1.9/modules/ROOT/pages/database-connector-select.adoc
```diff
@@ -0,0 +1,184 @@
...
+The combination limits the total amount of information that is retrieved (*Max rows* value) and guarantees that the data is returned from the database over the network in smaller chunks (*Fetch size* value):
```
kbostrom, line 127:
```suggestion
The combination of maximum rows and fetch sizes limits the total amount of information retrieved and guarantees that the data is returned from the database over the network in smaller chunks:
```

db/1.9/modules/ROOT/pages/database-connector-select.adoc
```diff
@@ -0,0 +1,184 @@
...
+
+. In your Studio flow, select the *Select* operation.
+. In the operation configuration screen, set the *SQL Query Text* field to `select * from some_table`.
+. Set the *Advanced* tab, set the *Fetch size* field to `200`, and the *Max rows* field to `1000`.
```
kbostrom:
```suggestion
. Click the *Advanced* tab, set the *Fetch size* field to `200`, and set the *Max rows* field to `1000`.
```

db/1.9/modules/ROOT/pages/database-connector-select.adoc
```diff
@@ -0,0 +1,184 @@
...
+
+== Configure the Query Timeout Fields
+
+The following factors often cause delays in query execution:
+
+* An inefficient query, such as one having improper indexing that iterates over many rows
+* A busy RDBMS or network
+* A lock contention
+
+To avoid timeouts when executing queries configure the *Query timeout* and *Query time unit* fields. All Database Connector operations support setting a timeout.
```
kbostrom: comma
```suggestion
To avoid timeouts when executing queries, configure the *Query timeout* and *Query time unit* fields. All Database Connector operations support setting a timeout.
```

db/1.9/modules/ROOT/pages/database-connector-select.adoc
```diff
@@ -0,0 +1,184 @@
...
+
+The following example shows how to set a timeout for the Select operation:
+
+. In your Studio flow, select the *Select* operation.
+. In the operation configuration screen, set the *SQL Query Text* field to `select * from some_table`.
+. In the *Advanced* tab, set the *Query timeout* field to the minimum amount of time before the JDBC driver attempts to cancel a running statement, for example `0`.
```
kbostrom: comma
```suggestion
. In the *Advanced* tab, set the *Query timeout* field to the minimum amount of time before the JDBC driver attempts to cancel a running statement, for example, `0`.
```

db/1.9/modules/ROOT/pages/database-connector-select.adoc
```diff
@@ -0,0 +1,184 @@
...
+. Set the *Query timeout unit* field to a time unit that qualifies the *Query timeout*, for example `SECONDS`.
```
kbostrom: comma
```suggestion
. Set the *Query timeout unit* field to a time unit that qualifies the *Query timeout*, for example, `SECONDS`.
```

kbostrom (dismissed):
CE per request

julodiaz (approved):

valkyrie69 (approved):

PR #1262, opened 2021-02-04
AMOD-3: Fix the documentation referring to the timeout

giss-ignacio:
Proposed fix for the documentation referring to the timeout and the aggregation complete route in the size based aggregation module.

Using a size based aggregator with a maxSize and a timeout. With a For Each scope a payload is sent that has a quantity X of elements that is bigger than maxSize but not a multiple of maxSize.
After giving timeout, the aggregation complete route is not executed, so the last (X mod maxSize) elements are not processed.

The **Async Versus Sync Aggregations** segment correctly documents the flows that can be made. But the description of the timeout says that the aggregation is considered complete when the timeout is reached, when it is not. Therefore, the Aggregation Complete Route is not executed when it times out.


aggregators/1.0/modules/ROOT/pages/aggregators-module-reference.adoc
```diff
@@ -70,10 +70,10 @@ This example aggregates elements until a pre-defined size is reached, executing
...
-| Timeout | Number |  A maximum time to wait for the aggregation to complete. If the timeout is reached before the total number of elements is equal to max size, the aggregation is considered complete. A value of `0` is not supported to avoid  timing out the group constantly. | -1(UNLIMITED) |
+| Timeout | Number |  A maximum time to wait for the aggregation to complete. If the timeout is reached before the total number of elements is equal to max size, the aggregation is not considered complete. A value of `0` is not supported to avoid  timing out the group constantly. | -1(UNLIMITED) |
 | Timeout unit | Time Unit | The time unit in which to measure the timeout. |  SECONDS |
 | Object Store | Object Store |  Either a name to reference a global object store or a definition of a private object store where the aggregated elements will be stored. |  Default Object Store partition. |
-| Aggregation Complete Route | Route | Components chain to execute once the aggregation is complete. | | x
+| Aggregation Complete Route | Route | Components chain to execute once the aggregation is complete. The Aggregation Complete Route will not execute if the timeout is reached. | | x
```
dhirsch1: Suggestion: "does not execute"

aggregators/1.0/modules/ROOT/pages/aggregators-module-reference.adoc
```diff
@@ -70,10 +70,10 @@ This example aggregates elements until a pre-defined size is reached, executing
 | Name | String | The name of the aggregator. It can later be referenced by an aggregator listener. || x
 | Content | Expression | The expression that defines what to aggregate. The result of the evaluation is the value stored in the aggregation. | #[payload] |
 | Max Size | Number | The total number of elements to be aggregated before considering the aggregation complete. | | x
-| Timeout | Number |  A maximum time to wait for the aggregation to complete. If the timeout is reached before the total number of elements is equal to max size, the aggregation is considered complete. A value of `0` is not supported to avoid  timing out the group constantly. | -1(UNLIMITED) |
+| Timeout | Number |  A maximum time to wait for the aggregation to complete. If the timeout is reached before the total number of elements is equal to max size, the aggregation is not considered complete. A value of `0` is not supported to avoid  timing out the group constantly. | -1(UNLIMITED) |
```
dhirsch1:
```suggestion
| Timeout | Number |  A maximum time to wait for the aggregation to complete. If the timeout is reached before the total number of elements is equal to max size, the aggregation is not considered complete. A value of `0` is not supported to avoid timing out the group constantly. | -1(UNLIMITED) |
```

kevintroller (dismissed):

dhirsch1 (approved):

estebanwasinger (approved):

PR #1287, opened 2021-02-16
Amazon SQS 5.10

dhirsch1:


xidex:
@dhirsch1 @sathya0 what should be the content of migration guide? 
In our case we had version 5.5.0 deprecated so the migration guide was created for that version to migrate clients either back to 5.4.3 or forward to 5.6.0+. 
But since then every release a newer version was added to this document. I would like to verify with you guys if it's correct.

sathya0:
@xidex We usually do the Migrate topic for major releases only. Not sure why there are upgrade updates for the minor releases for this connector, but since there are we will have one for 5.10.

amazon-sqs/5.10/modules/ROOT/pages/amazon-sqs-connector-upgrade-migrate.adoc
```diff
@@ -0,0 +1,101 @@
...
+[%header,cols="50a,50a"]
+|===
+|From Version | To Version
+|5.x |5.10
+|===
+
+== What's Changed
+=== 5.10
+
+Amazon SQS Connector now supports the addition of SQS endpoint which can be used for utilizing Amazon VPC.
```
dhirsch1:
```suggestion
Amazon SQS Connector now supports the addition of an SQS endpoint that you can use with Amazon Virtual Public Cloud (VPC).
```

valkyrie69 (approved):

PR #1298, opened 2021-02-22
Change verbiage for vm:consume operation

sairamkodipyaka:
Current verbiage for vm:consume reads 'The flow in which the content is published cannot contain a Listener (<vm:listener>) source'. Here 'content is published' is slightly misleading as the topic is about consuming the message.


kevintroller:
Thanks for reporting the update @sairamkodipyaka I'm closing this PR, since we addressed this change on PR https://github.com/mulesoft/docs-connectors/pull/1300

PR #1301, opened 2021-02-22
Added latest kafka connector reference changes

EmilMarian:


EmilMarian:
@valkyrie69 please disregard this PR, as it is not correct. Please review the following one instead. https://github.com/mulesoft/docs-connectors/pull/1305

PR #1339, opened 2021-03-05
Update sap-connector-reference.adoc (details about SAP and Bitronix)

ca-stefan-cordes:
Refer to "what to do" in case using Bitronix and SAP


valkyrie69:
> Refer to "what to do" in case using Bitronix and SAP



> Refer to "what to do" in case using Bitronix and SAP

Hi Stefan, 
It looks like you are requesting that we document something that will be implemented in the future. We don't document future features though.  Thank you.

ca-stefan-cordes:
No, this is the instruction how to workaround the limitation.

Instead of writing limitation "abc" is not working you should give a hint what do to against that limitation.

PR #1341, opened 2021-03-08
"...limitations." (Finishing the sentence.)

davidvancina-mulesoft:
It would be good to say, if possible, what those limitations are. Approver: Perhaps open a companion Jira for that, as a follow-up?


valkyrie69 (approved):

valkyrie69:
> It would be good to say, if possible, what those limitations are. Approver: Perhaps open a companion Jira for that, as a follow-up?

Hi David! I filed a Jira to enhance the docs to specify what those limitations are. Thank you for your contribution!

PR #1368, opened 2021-03-19
Update amazon-sns-connector-upgrade-migrate.adoc

sathya0:


amazon-sns/4.5/modules/ROOT/pages/amazon-sns-connector-upgrade-migrate.adoc
```diff
@@ -1,19 +1,19 @@
...
+= Upgrade Amazon SNS Connector to Version 4.5
 :page-aliases: connectors::amazon/amazon-sns-connector-upgrade-migrate.adoc
 
 == Supported Upgrade Paths
 
 [%header,cols="50a,50a"]
 |===
 |From Version | To Version
-|4.4.0 |4.5.0
+|4.3.1 |4.5.0
```
dhirsch1: I would say 4.5.x instead of 4.5.0 because we don't always need to upgrade this topic for a minor release and it usually applies to the patch releases for the minor release.

dhirsch1 (approved):

PR #1385, opened 2021-03-29
oracle ebs 5.2

sathya0:


oracle-ebs/5.2/modules/ROOT/pages/index.adoc
```diff
@@ -0,0 +1,397 @@
+= Oracle E-Business Suite 12.1 Connector
```
valkyrie69: Lets make the titles consistent use EBS.

oracle-ebs/5.2/modules/ROOT/pages/oracle-ebs-connector-reference.adoc
```diff
@@ -0,0 +1,2656 @@
+= Oracle EBS 12.1 Connector Reference
+:page-aliases: connectors::oracle/oracle-ebs-connector-reference.adoc
+
+Support Category: https://www.mulesoft.com/legal/versioning-back-support-policy#anypoint-connectors[Premium]
+
+Oracle EBS 12.1 Connector Version 5.2
```
valkyrie69:
```suggestion
Anypoint Connector for Oracle EBS 12.1 (Oracle EBS 12.1 Connector) Version 5.2
```

valkyrie69: Normally, we would remove this line but I think in this case it helps avoid the confusion caused by the fact 12.1 is part of the connector name. 
The title would normally have the connector version but in this case I think it would be confusing to have: Oracle EBS 12.1 Connector 5.2 Reference. 
What do you think?

oracle-ebs/5.2/modules/ROOT/pages/oracle-ebs-connector-reference.adoc
```diff
@@ -0,0 +1,2656 @@
...
+
+Release Notes: xref:release-notes::connector/oracle-ebs-connector-release-notes-mule-4.adoc[Oracle EBS 12.1 Connector Release Notes]
```
valkyrie69: Delete the link to the release notes since it's in the info box now.

oracle-ebs/5.2/modules/ROOT/pages/oracle-ebs-connector-reference.adoc
sathya0: Hmm, not sure what you mean by connector version, but I think we should keep the 12.1 just because there is a 12.2 version as well so it would be more clear which docs the user is looking at. But if you mean the version 5.2, then yes I think we can remove that.

valkyrie69 (approved):

PR #1429, opened 2021-04-22
Added a note to call out incompatibility with Windows

amitfernandes:


dhirsch1:
Per Amit, this PR is not needed.

PR #1452, opened 2021-05-06
Processing Events with Salesforce Connector - Topic Revamp

dhirsch1:
IMPORTANT: This is the original topic that was done for v10.10. Since the connector is now on v10.11, the updated file needs to be copied to this PR: https://github.com/mulesoft/docs-connectors/pull/1486

### DIFF
--- a/salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
+++ b/salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
@@ -1,19 +1,18 @@
 = Processing Events With Salesforce Connector 10.10
 :page-aliases: connectors::salesforce/salesforce-connector-processing-events.adoc
 
-You can use Anypoint Connector for Salesforce (Salesforce Connector) as an inbound connector to stream data from Salesforce into your application. To use the connector this way, use one of these source operations:
+You can use Anypoint Connector for Salesforce (Salesforce Connector) as an inbound connector to stream data from Salesforce into your application. To use the connector this way, use one of these sources:
 
 * *Subscribe topic listener*
 * *Subscribe channel listener*
 * *Replay topic listener*
 * *Replay channel listener*
 
-[NOTE]
-When deploying apps on a distributed environment such as a runtime cluster, the source operation must run on the primary node. If the same source runs on multiple nodes, it causes the consumption of duplicate events.
+When deploying apps in a distributed environment such as a runtime cluster, the source must run on the primary node. Running the source on multiple nodes causes duplicate event consumption.
 
 Each event that travels through your flow contains information about the Salesforce data that changed, including how the data changed and when the change occurred.
 
-Salesforce stores events for 24 hours (or 72 hours for high-volume events). A subscriber to a topic or channel can retrieve events related to that topic or channel during the 24-hour retention window. After the retention window ends, the subscriber can retrieve newer events that have not yet expired.
+Salesforce stores events for 24 hours (or 72 hours for high-volume events). A subscriber to a topic or channel can retrieve events related to that topic or channel during this retention window, after which the subscriber can retrieve only newer events that have not yet expired.
 
 Salesforce assigns each broadcast event a numeric ID that is unique to the organization and channel. IDs are incremented, but not necessarily consecutively; for example, the event following event `999` might be event `1025`. Salesforce does not reuse the IDs of deleted events.
 
@@ -45,11 +44,11 @@ When a Mule app starts for the first time, the connector creates an object store
 * For each message that processes successfully, the connector stores the associated replay ID.
 * For each message that fails, the connector stores a number that represents the lowest replay ID for which the processing failed.
 
-These two structures ensure that no message is processed twice and that when the application restarts, it can reprocess the failed messages.
+Saving this change_data_capture in an object store ensures that no message is processed twice and that when the application restarts, it can reprocess any failed messages.
 
 The IDs are stored in the object store for 72 hours, after which they are automatically removed.
 
-For each message that comes through a topic or streaming channel to which the connector is subscribed, the connector updates the object store information using up to four transactions.
+For each message that comes through a topic or streaming channel to which the connector is subscribed, the connector uses up to four transactions to update the object store information.
 
 If you have multiple workers deployed in CloudHub, all Mule instances process each message when using the *Replay Topic* operation. To avoid multiple Mule instances processing the same message, you can do one of the following:
 
@@ -60,11 +59,11 @@ If you have multiple workers deployed in CloudHub, all Mule instances process ea
 
 To enable the *Resume from the Last Replay Id* functionality for the *Replay Topic Listener* and *Replay Channel Listener* sources, the connector must store information about the events that were successfully processed by the flow in which the source is used. The connector uses an object store to store that information.
 
-The way that the object store stores the necessary data for the *Resume from the Last Replay Id* varies depending on the environment:
+The way that the object store stores the necessary data for the *Resume from the Last Replay Id* operation varies depending on the environment:
 
 * When a Mule app runs locally, the connector creates a persistent store in the file system.
 +
-If the application stops and restarts, or if a connection issue triggers a reconnection, the data saved in the object store enables the connector to continue processing messages at the point where it previously stopped.
+If the application stops and restarts, or if a connection issue triggers a reconnection, the data saved in the object store enables the connector to continue processing messages at the point where it stopped.
 * When a Mule app runs in a cluster with the nodes either on the same machine or on different machines, Mule uses a distributed in-memory object store:
 ** If the primary node (the one consuming the messages) shuts down, the node that takes over for the primary node does not reprocess the same messages. Instead, the node continues from where the previous node stopped processing the messages.
 ** If the cluster stops, the data from the object store is lost. When the cluster restarts, the configuration at the source level determines whether the connector reprocesses the messages.
--- a/salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
+++ b/salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
@@ -24,7 +24,7 @@ To process events with Salesforce Connector, ensure that you have:
 * Access to the Salesforce target resource
 * The required Salesforce Streaming API permissions enabled in your organization
 
-Before you can receive events for Salesforce changes associated with a topic, create a topic using Salesforce Connector if the topic doesn't already exist.
+To receive events for Salesforce changes associated with a topic, create a new topic or use an existing one.
 
 [[objectstoreusage]]
 == Object Store Usage in Events
@@ -46,6 +46,8 @@ When a Mule app starts for the first time, the connector creates an object store
 * For each message that fails, the connector stores a number that represents the lowest replay ID for which the processing failed.
 
 These two structures ensure that no message is processed twice and that when the application restarts, it can reprocess the failed messages.
+
+The IDs are stored in the object store for 72 hours, after which they are automatically removed.
 
 For each message that comes through a topic or streaming channel to which the connector is subscribed, the connector updates the object store information using up to four transactions.
 
--- a/salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
+++ b/salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
@@ -15,15 +15,16 @@ Each event that travels through your flow contains information about the Salesfo
 
 Salesforce stores events for 24 hours (or 72 hours for high-volume events). A subscriber to a topic or channel can retrieve events related to that topic or channel during the 24-hour retention window. After the retention window ends, the subscriber can retrieve newer events that have not yet expired.
 
-Salesforce assigns each broadcast event a numeric ID. IDs are incremented, but not necessarily by 1 for each consecutive event. For example, the event following the event with ID 999 can have an ID of 1025. A broadcast event ID is unique for the organization and channel. Salesforce does not reuse the IDs of deleted events.
+Salesforce assigns each broadcast event a numeric ID. IDs are incremented, but not necessarily by 1 for each consecutive event. For example, the event following the event with ID `999` can have an ID of `1025`. A broadcast event ID is unique for the organization and channel. Salesforce does not reuse the IDs of deleted events.
 
 == Prerequisites
 
-To process events with Salesforce Connector, ensure that you:
+To process events with Salesforce Connector, ensure that you have:
 
-* Have access to the Salesforce target resource.
-* Have the required Salesforce Streaming API permissions enabled in your organization.
-* Create a topic using Salesforce Connector, if it doesn't already exist, before you can receive events for Salesforce changes associated with the topic.
+* Access to the Salesforce target resource.
+* The required Salesforce Streaming API permissions enabled in your organization.
+
+Before you can receive events for Salesforce changes associated with a topic, create a topic using Salesforce Connector, if the topic doesn't already exist.
 
 [[objectstoreusage]]
 == Object Store Usage in Events
@@ -48,14 +49,14 @@ These two structures ensure that no message is processed twice and that when the
 
 For each message that comes through a topic or streaming channel to which the connector is subscribed, the connector updates the object store information using up to four transactions.
 
-If you have multiple workers deployed in CloudHub, all Mule instances process each message when using the *Replay Topic* operations. To avoid multiple Mule instances processing the same message, you can do one of the following:
+If you have multiple workers deployed in CloudHub, all Mule instances process each message when using the *Replay Topic* operation. To avoid multiple Mule instances processing the same message, you can do one of the following:
 
 * Modify your app to filter out duplicate messages.
 * Designate a single worker as a receiver and then use a shared queue to pass the messages to a different worker to process the message.
 
 === Resume from the Last Replay ID
 
-To enable the *Resume from the Last Replay Id* functionality for the *Replay Topic Listener* and *Replay Channel Listener* sources, the connector must store information about the events that were successfully processed by the flow in which the source is used. The connector uses Object Store to store that information.
+To enable the *Resume from the Last Replay Id* functionality for the *Replay Topic Listener* and *Replay Channel Listener* sources, the connector must store information about the events that were successfully processed by the flow in which the source is used. The connector uses Object Store *(Reviewers, is this always the Mule object store or can it be another object store?)* to store that information.
 
 The way Object Store stores the necessary data for the *Resume from the Last Replay Id* varies depending on the environment:
 
@@ -85,11 +86,11 @@ Work with topic events by creating a topic, subscribing to a topic, and replayin
 
 When you create a topic, the connector creates a `PushTopic`, which is a special object in Salesforce that binds a name (in this case, the topic's name) and a Salesforce Object Query Language (SOQL) query together. After you create a topic, you can then subscribe to it by its name.
 
-You can use either the *Create* (`create`) or *Publish topic* (`publish-topic`) operation to create a topic. The following example uses the `publish-topic` operation to create a topic:
+You can use either the *Create* (`create`) operation or *Publish topic* (`publish-topic`) operation to create a topic. The following example uses the `publish-topic` operation to create a topic:
 
 `<sfdc:publish-topic name="AccountUpdates" query="SELECT Id, Name FROM Account"/>`
 
-Alternatively, you can create a topic in Salesforce by executing code such as this code from an *Enter Apex Code* window, which is accessible through the system logs:
+Alternatively, you can create a topic in Salesforce by executing code from an *Enter Apex Code* window, which is accessible through the system logs, for example:
 
 [source,text,linenums]
 ----
@@ -142,9 +143,9 @@ If you specify either the `ALL` replay option or `ONLY_NEW` replay option, the `
 
 The *Resume from the Last Replay Id* checkbox enables you to specify an automatic replay of stored events based on the Replay ID of the last event processed by the connector. You can use this functionality when the connector stops listening, such as a during a server shutdown or dropped connection. If the stored Replay ID is outside the 24-hour retention period, the replay option determines what events to replay.
 
-To support the *Resume from Last Replay ID* feature, the connector uses a persistent object store to keep different details regarding the processed messages. This feature is enhanced in Salesforce Connector 10.x to reduce message loss probability and to avoid processing duplicate messages. For more details about how Object Store is used, check the <<objectstoreusage,Object Store Usage>> section.
+To support the *Resume from Last Replay ID* feature, the connector uses a persistent object store to keep different details regarding the processed messages. This feature is enhanced in Salesforce Connector 10.x and later to reduce message loss probability and to avoid processing duplicate messages. For more details about how Object Store is used, check the <<objectstoreusage,Object Store Usage>> section.
 
-In the following XML example, the *Replay topic listener* operation acts like an inbound endpoint for the Logger component message:
+In the following XML example, the *Replay topic listener* operation acts like an inbound endpoint for the *Logger* component message:
 
 [source,xml,linenums]
 ----
@@ -185,7 +186,7 @@ You can also create a streaming channel by using either the connector *Create* o
 [[streamsub]]
 === Subscribe to a Streaming Channel
 
-After you create a streaming channel, you can start receiving events by subscribing to the channel. The `subscribe-channel-listener` source acts like an inbound endpoint. In this example, every time a subscription to `/u/TestStreaming` receives an event, it executes the rest of the flow and logs a message at the INFO level:
+After you create a streaming channel, you can start receiving events by subscribing to the channel. The *Subscribe channel listener* (`subscribe-channel-listener`) source acts like an inbound endpoint. In this example, every time a subscription to `/u/TestStreaming` receives an event, it executes the rest of the flow and logs a message at the INFO level:
 
 [source,xml,linenums]
 ----
@@ -204,7 +205,7 @@ For more information, see https://developer.salesforce.com/docs/atlas.en-us.chan
 [[channelrep]]
 === Replay Messages From a Streaming Channel
 
-A streaming channel can replay notifications. The *Replay channel listener* input source acts as an inbound endpoint. You can use it as shown in the following example:
+A streaming channel can replay notifications. The *Replay channel listener* input source acts as an inbound endpoint, for example:
 
 [source,xml,linenums]
 ----
@@ -220,15 +221,15 @@ If you specify either the `ALL` replay option or the `ONLY_NEW` replay option, t
 
 The *Resume from the Last Replay Id* checkbox enables you to specify an automatic replay of stored events, based on the Replay ID of the last event processed by the connector. You can use this functionality when the connector stops listening, such as a during a server shutdown or dropped connection. If the stored Replay ID is outside the 24-hour retention period, the replay option determines what events to replay.
 
-To support the *Resume from Last Replay ID* feature, the connector uses a persistent object store to keep different details regarding the processed messages. This feature is enhanced in Salesforce Connector 10.x to reduce message loss probability and to avoid processing duplicate messages. For more details about how Object Store is used, check the <<objectstoreusage,Object Store Usage>> section.
+To support the *Resume from Last Replay ID* feature, the connector uses a persistent object store to keep different details regarding the processed messages. This feature reduce messages loss probability and avoids processing duplicate messages. For more details about how Object Store is used, check the <<objectstoreusage,Object Store Usage>> section.
 
 === Push Events to a Streaming Channel with No Listener
 
 Users can push events to a streaming channel, even if the channel does not have a listener for reading published events. Once a listener is started for the channel, Salesforce Streaming API pushes as many messages it can to the listener, based on the maximum daily limit.
 
-For example, given that the maximum number of delivered event notifications within a 24-hour period for a free Salesfoce organization is 10,000, suppose you publish 15,000 events to that channel. When Salesforce Connector subscribes to that channel, the Streaming API attempts to push 10,000 events, thus consuming the daily quota. The API will then attempt to push the remaining 5,000 events the next day before pushing any new events.
+For example, given that the maximum number of delivered event notifications within a 24-hour period for a free Salesforce organization is 10,000, suppose you publish 15,000 events to that channel. When Salesforce Connector subscribes to that channel, Streaming API attempts to push 10,000 events, thus consuming the daily quota. The API will then attempt to push the remaining 5,000 events the next day before pushing any new events.
 
-In this scenario, the connector streams the events one-by-one into the Mule app. If the app takes too long to process a message, the Salesforce Streaming API might instruct the connector to reconnect. If this happens, the Streaming API drops all of the unprocessed messages.
+In this scenario, the connector streams the events one-by-one into the Mule app. If the app takes too long to process a message, Streaming API might instruct the connector to reconnect. If this happens, Streaming API drops all of the unprocessed messages.
 
 You can avoid this situation by implementing a reliability pattern, as described in xref:mule-runtime::reliability-patterns.adoc[Reliability Patterns].
 
@@ -260,7 +261,7 @@ For more information about working with streaming channels, see <<createchannel,
 
 Salesforce enables you to push custom events to a specific streaming channel through the REST API. To do this, use https://workbench.developerforce.com/about.php[Workbench] or this connector.
 
-The following example uses the connector's `push-generic-event` operation to push custom events to the channel with the ID `0M6j0000000KyjBCAS`:
+The following example uses the connector's *Push generic event* (`push-generic-event`) operation to push custom events to the channel with the ID `0M6j0000000KyjBCAS`:
 
 [source,xml,linenums]
 ----
@@ -275,7 +276,7 @@ The following example uses the connector's `push-generic-event` operation to pus
 </flow>
 ----
 
-You can retrieve the channel ID from the response map of the `publish-streaming-channel` operation. Alternatively, you can retrieve the channel ID from the Salesforce page:
+You can retrieve the channel ID from the response map of the *Publish streaming channel* operation. Alternatively, you can retrieve the channel ID from the Salesforce page:
 
 . Log in to your Salesforce Developer Edition organization.
 . Under *All Tabs (+)*, select *Streaming Channels*.
--- a/salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
+++ b/salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
@@ -48,7 +48,7 @@ These two structures ensure that no message is processed twice and that when the
 
 For each message that comes through a topic or streaming channel to which the connector is subscribed, the connector updates the object store information using up to four transactions.
 
-Note: If you have multiple workers deployed in CloudHub, all Mule instances process each message when using the *Replay Topic* operations.
+NOTE: If you have multiple workers deployed in CloudHub, all Mule instances process each message when using the *Replay Topic* operations.
 
 To avoid multiple Mule instances processing the same message, you can do one of the following:
 
@@ -76,7 +76,7 @@ To persist data so that the app continues processing events when restarted, crea
 
 Avoid having two or more flows use a source that listens to events on the same channel or topic because the events will be consumed twice from your user quota. If you have a use case that requires processing the same events in two or more different flows, consider implementing that logic in a single flow instead.
 
-Having multiple sources listening to the same events and having the *Resume from the Last Replay Id* option enabled can lead to data corruption and potential loss of events.
+NOTE: Having multiple sources listening to the same events and having the *Resume from the Last Replay Id* option enabled can lead to data corruption and potential loss of events.
 
 == Work with Topic Events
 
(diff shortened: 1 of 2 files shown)


salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
```diff
@@ -9,88 +9,88 @@ You can use Anypoint Connector for Salesforce (Salesforce Connector) as an inbou
...
-To process events with Salesforce Connector, ensure you meet the following prerequisites:
+To process events with Salesforce Connector, ensure that you have:
+
+* Access to the Salesforce target resource
+* The required Salesforce Streaming API permissions enabled in your organization
 
-* You must have access to the Salesforce target resource.
-* Before you can receive events for Salesforce changes associated with a topic, you must create the topic, if it does not exist.
-* You must have the required Salesforce Streaming API permissions enabled in your organization.
+Before you can receive events for Salesforce changes associated with a topic, create a topic using Salesforce Connector if the topic doesn't already exist.
```
stefangretcan: The channel can also be created from Salesforce UI not necessary only with the mule connector.

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
```diff
@@ -9,88 +9,88 @@ You can use Anypoint Connector for Salesforce (Salesforce Connector) as an inbou
...
 
 [[objectstoreusage]]
-== Object Store Usage
+== Object Store Usage in Events
 
-Both Salesforce Connector and Mule use an object store to persist data for features such as automatic message replay and message redelivery:
+Both Salesforce Connector and Mule runtime engine (Mule) use an object store to persist data for features such as automatic message replay and message redelivery:
```
stefangretcan, line 31: We should only mention that the Salesforce Connector uses an object store to persist data for feature such as...

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
```diff
@@ -9,88 +9,88 @@ You can use Anypoint Connector for Salesforce (Salesforce Connector) as an inbou
...
-When a Mule app starts for the first time, the connector creates an object store that is used to save data related to messages that were processed successfully or that failed:
+When a Mule app starts for the first time, the connector creates an object store that saves data:
 
-* For each message that was processed successfully, the information stored consists of a replay ID associated with the message.
-* For each message that failed, the information stored consists of a number that represents the lowest replay ID for which the processing failed.
+* For each message that processes successfully, the connector stores the replay ID associated with the message.
+* For each message that fails, the connector stores a number that represents the lowest replay ID for which the processing failed.
 
-These two structures ensure that no message is processed twice and that when the application is restarted, it can reprocess the failed messages.
+These two structures ensure that no message is processed twice and that when the application restarts, it can reprocess the failed messages.
```
stefangretcan: + These information is stored in the object store for 72 hours after which they are automatically removed.

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
```diff
@@ -9,88 +9,88 @@ You can use Anypoint Connector for Salesforce (Salesforce Connector) as an inbou
...
 
-=== Streaming Channel with No Listener
+== Multiple Flows That Use the Same Source
 
-Users can push events to a streaming channel, even if the channel does not have a listener for reading published events. Once a listener is started for the channel, Salesforce Streaming API pushes as many messages it can to the listener, based on the maximum daily limit.
+Avoid having two or more flows use a source that listens to events on the same channel or topic, because the events will be consumed twice from your user quota. If you have a use case that requires processing the same events in two or more different flows, consider implementing that logic in a single flow instead.
 
-For example, given that the maximum number of delivered event notifications within a 24-hour period for a free Salesfoce organization is 10,000, suppose you publish 15,000 events to that channel. When Salesforce Connector subscribes to that channel, the Streaming API attempts to push 10,000 events, thus consuming the daily quota. The API will then attempt to push the remaining 5,000 events the next day before pushing any new events.
```
stefangretcan, line 78: I think this was a good example/explanation that was just removed and not refactored. Let's make sure we do not loose information.

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
```diff
@@ -9,88 +9,88 @@ You can use Anypoint Connector for Salesforce (Salesforce Connector) as an inbou
...
+WARNING: Having multiple sources listening to the same events and having the *Resume from the Last Replay Id* option enabled can lead to data corruption and the potential loss of events.
 
-In this scenario, the connector streams the events one-by-one into the Mule app. If the app takes too long to process a message, the Salesforce Streaming API might instruct the connector to reconnect. If this happens, the Streaming API drops all of the unprocessed messages.
```
stefangretcan, line 80: Was this removed?

stefangretcan (changes requested):
We might need Alex's approval too

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
alexandruatomei9: Yes I would rephrase this as "create a topic or use an already existing one."

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
alexandruatomei9: No @stefangretcan it is important to mention that the runtime uses OS too because if there are only 10 transactions per second allowed, then there might be other components also consuming some of those transactions.

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
alexandruatomei9: I agree this is important, might be worth adding it to the doc.

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
alexandruatomei9: @stefangretcan this was moved to another section. You can still find it on the right side.

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
alexandruatomei9: No

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
```diff
@@ -9,88 +9,90 @@ You can use Anypoint Connector for Salesforce (Salesforce Connector) as an inbou
 * *Replay channel listener*
 
 [NOTE]
-When deploying apps on a distributed environment (such as a runtime cluster), the source operation must run on the primary node. If the same source runs on multiple nodes, it causes duplicate event consumption.
+When deploying apps on a distributed environment such as a runtime cluster, the source operation must run on the primary node. If the same source runs on multiple nodes, it causes the consumption of duplicate events.
```
kbostrom: Is this a NOTE? How much of it? Does it really need to be? See the style ref. for ways to evaluate.

kbostrom: "in an environment" rather than "on." I suggest using prepositions in tech doc as they would be in regular doc: on a page, in a field, in an environment, on a node, and so on.
Let me know if there is a technical or industry-standard reason not to do so.
I am working on a proposal to present the team re. prepositions with UI elements.

kbostrom: tiny cleanup re. source...nodes...it:
```suggestion
When deploying apps in a distributed environment such as a runtime cluster, the source operation must run on the primary node. Running the source operation on multiple nodes causes <destructive? unnecessary?> duplicate event consumption.
```

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
```diff
@@ -9,88 +9,90 @@ You can use Anypoint Connector for Salesforce (Salesforce Connector) as an inbou
...
+
+Each event that travels through your flow contains information about the Salesforce data that changed, including how the data changed and when the change occurred.
+
+Salesforce stores events for 24 hours (or 72 hours for high-volume events). A subscriber to a topic or channel can retrieve events related to that topic or channel during the 24-hour retention window. After the retention window ends, the subscriber can retrieve newer events that have not yet expired.
```
kbostrom:
```suggestion
Salesforce stores events for 24 hours (or 72 hours for high-volume events). A subscriber to a topic or channel can retrieve events related to that topic or channel during this retention window, after which the subscriber can  retrieve newer events that have not yet expired.
```
...with these questions:
- Does the second sentence apply to the high-volume period as well?
- Are we saying that the subscriber cannot access any subsequent events the current event-retention window expires? Is that a good thing? What if the newer events have expired...does the user miss them altogether or have they been resolved? (If this is just me not understanding what your audience will, then ignore it...after making sure that this text is as clear as it can be.)

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
```diff
@@ -9,88 +9,90 @@ You can use Anypoint Connector for Salesforce (Salesforce Connector) as an inbou
...
+
+Salesforce assigns each broadcast event a numeric ID. IDs are incremented, but not necessarily by 1 for each consecutive event. For example, the event following the event with ID `999` can have an ID of `1025`. A broadcast event ID is unique for the organization and channel. Salesforce does not reuse the IDs of deleted events.
```
kbostrom: Why are we now specifying "broadcast" along with "event"? Should we have earlier?

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
```diff
@@ -9,88 +9,90 @@ You can use Anypoint Connector for Salesforce (Salesforce Connector) as an inbou
 * *Replay channel listener*
```
kbostrom, line 10: Fix title "= Processing Events With Salesforce Connector 10.10" to use "Process" and lowercase "with."
https://capitalizemytitle.com/#

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
kbostrom: Simpler?
```suggestion
Salesforce assigns each broadcast event a numeric ID that is unique to the organization and channel. IDs are incremented, but not necessarily consecutively; for example, the event following event `999` might be event `1025`. Salesforce does not reuse the IDs of deleted events.
```

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
```diff
@@ -9,88 +9,90 @@ You can use Anypoint Connector for Salesforce (Salesforce Connector) as an inbou
...
-To process events with Salesforce Connector, ensure you meet the following prerequisites:
+To process events with Salesforce Connector, ensure that you have:
+
+* Access to the Salesforce target resource
+* The required Salesforce Streaming API permissions enabled in your organization
 
-* You must have access to the Salesforce target resource.
-* Before you can receive events for Salesforce changes associated with a topic, you must create the topic, if it does not exist.
-* You must have the required Salesforce Streaming API permissions enabled in your organization.
+To receive events for Salesforce changes associated with a topic, create a new topic or use an existing one.
```
kbostrom: This is subjective, but I like the use of "either" to prepare the reader for two options:
```suggestion
To receive events for Salesforce changes associated with a topic, either create a new topic or use an existing one.
```

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
```diff
@@ -9,88 +9,90 @@ You can use Anypoint Connector for Salesforce (Salesforce Connector) as an inbou
...
 
 [[objectstoreusage]]
-== Object Store Usage
+== Object Store Usage in Events
```
kbostrom, line 29: Do we want "Usage" or "Use"?

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
```diff
@@ -9,88 +9,90 @@ You can use Anypoint Connector for Salesforce (Salesforce Connector) as an inbou
...
 * The free version of Object Store for CloudHub deployments of Mule apps has a limit of 10 transactions per second.
 
-For more information about object store versions, see https://docs.mulesoft.com/object-store/#object-store-notes[Object Store Notes].
-
+For more information about Object Store versions, see https://docs.mulesoft.com/object-store/#object-store-notes[Object Store Notes].
 
 === Replay Topic and Replay Channel Listener Operations
 
-The *Replay topic* and *Replay channel listener* operations have the option to continue from the last replay ID they received before restarting the application.
+The *Replay topic* and *Replay channel listener* operations can continue from the last replay ID they received before restarting the application.
```
kbostrom, line 40: option vs. ability: Do we mean "might" instead of "can"?

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
```diff
@@ -9,88 +9,90 @@ You can use Anypoint Connector for Salesforce (Salesforce Connector) as an inbou
...
 
-When a Mule app starts for the first time, the connector creates an object store that is used to save data related to messages that were processed successfully or that failed:
+When a Mule app starts for the first time, the connector creates an object store that saves data:
 
-* For each message that was processed successfully, the information stored consists of a replay ID associated with the message.
-* For each message that failed, the information stored consists of a number that represents the lowest replay ID for which the processing failed.
+* For each message that processes successfully, the connector stores the replay ID associated with the message.
```
kbostrom:
```suggestion
* For each message that processes successfully, the connector stores the associated replay ID.
```

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
```diff
@@ -9,88 +9,90 @@ You can use Anypoint Connector for Salesforce (Salesforce Connector) as an inbou
...
+* For each message that fails, the connector stores a number that represents the lowest replay ID for which the processing failed.
```
kbostrom: "a number..." versus the actual replay ID, as implied by the peer list item?
```suggestion
* For each message that fails, the connector stores the lowest replay ID for which the processing failed.
```
I'm confused: if something fails, it fails once, at the first instance that it cannot process. Isn't this automatically the lowest/first ID for which it fails: do we need to say "lowest"? Are there other failures after the first?

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
```diff
@@ -9,88 +9,90 @@ You can use Anypoint Connector for Salesforce (Salesforce Connector) as an inbou
...
 
-These two structures ensure that no message is processed twice and that when the application is restarted, it can reprocess the failed messages.
+These two structures ensure that no message is processed twice and that when the application restarts, it can reprocess the failed messages.
```
kbostrom: "structures" the best word?

kbostrom: "the" failed messages or "any" failed messages?

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
```diff
@@ -9,88 +9,90 @@ You can use Anypoint Connector for Salesforce (Salesforce Connector) as an inbou
...
 
-For each message that comes through a topic or streaming channel to which the connector is subscribed, the connector updates the object store information using up to four transactions.
+The IDs are stored in the object store for 72 hours, after which they are automatically removed.
```
kbostrom: Is this related to Line 16?

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
```diff
@@ -9,88 +9,90 @@ You can use Anypoint Connector for Salesforce (Salesforce Connector) as an inbou
...
 
-[NOTE]
-====
-If you have multiple workers deployed in CloudHub, all Mule instances process each message when using the *Replay Topic* or *Replay Channel* operations.
+For each message that comes through a topic or streaming channel to which the connector is subscribed, the connector updates the object store information using up to four transactions.
```
kbostrom: Does this mean that the object store information can be updated only four times, or that it requires one to four transactions for the connector to update the information?

kbostrom: (I'm feeling either picky or dense today: sorry if these are stoopid questions....)

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
```diff
@@ -9,88 +9,90 @@ You can use Anypoint Connector for Salesforce (Salesforce Connector) as an inbou
...
 
-To avoid multiple Mule instances processing the same message, you can do one of the following:
+If you have multiple workers deployed in CloudHub, all Mule instances process each message when using the *Replay Topic* operation. To avoid multiple Mule instances processing the same message, you can do one of the following:
```
kbostrom: To what does "when using" the operation refer: workers, CloudHub, instances, or message? What is using the operation? Can you reorder the wording to clarify?

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
```diff
@@ -9,88 +9,90 @@ You can use Anypoint Connector for Salesforce (Salesforce Connector) as an inbou
...
 * Modify your app to filter out duplicate messages.
 * Designate a single worker as a receiver and then use a shared queue to pass the messages to a different worker to process the message.
-====
 
 === Resume from the Last Replay ID
 
-To enable the *Resume from the Last Replay Id* functionality for the *Replay Topic Listener* and *Replay Channel Listener* source operations, the connector must store some information about the events that were successfully processed by the flow in which the source operation is used. The connector uses Object Store to store that information.
-
-The way Object Store is used to store the necessary data for the *Resume from the Last Replay Id* varies depending on the environment:
+To enable the *Resume from the Last Replay Id* functionality for the *Replay Topic Listener* and *Replay Channel Listener* sources, the connector must store information about the events that were successfully processed by the flow in which the source is used. The connector uses an object store to store that information.
```
kbostrom: Why boldface functionality and source names? Are these operations?

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
```diff
@@ -9,88 +9,90 @@ You can use Anypoint Connector for Salesforce (Salesforce Connector) as an inbou
...
-* When running the application in a cluster, either with the nodes on the same machine, or with the nodes on different machines, a distributed in-memory object store is used. This way, if the primary node (the one consuming the messages) shuts down, the node that takes over for the primary node will not consume the same messages again, but will continue from where the previous node stopped processing the messages. +
-If the cluster is stopped, the data from the object store is lost, so when the cluster restarted the messages might be reprocessed again depending on the configuration at the source level. +
-If you need data to be persisted so that the application can continue processing events when restarted, the object store can be configured to use a database as the persistence layer. For more information, see xref:mule-runtime::creating-and-managing-a-cluster-manually.adoc[Creating and Managing a Cluster]).
-* When running the application in CloudHub, if you are using multiple workers, all of the workers receive and process the events.
+The way that the object store stores the necessary data for the *Resume from the Last Replay Id* varies depending on the environment:
```
kbostrom:
```suggestion
The way that the object store stores the necessary data for *Resume from the Last Replay Id* depends on the environment:
```

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
```diff
@@ -9,88 +9,90 @@ You can use Anypoint Connector for Salesforce (Salesforce Connector) as an inbou
...
 
-=== Multiple Flows That Use the Same Source
-
-It is best that you do not have two or more flows using a source that listens to events on the same channel or topic because the events will be consumed twice from your user quota. If you have a use case that requires processing the same events in two or more different flows, consider implementing that logic in a single flow instead.
+* When a Mule app runs locally, the connector creates a persistent store in the file system.
++
+If the application stops and restarts, or if a connection issue triggers a reconnection, the data saved in the object store enables the connector to continue processing messages at the point where it previously stopped.
```
kbostrom: Do we need "previously"?

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
```diff
@@ -9,88 +9,90 @@ You can use Anypoint Connector for Salesforce (Salesforce Connector) as an inbou
...
+* When a Mule app runs in a cluster with the nodes either on the same machine or on different machines, Mule uses a distributed in-memory object store:
+** If the primary node (the one consuming the messages) shuts down, the node that takes over for the primary node does not reprocess the same messages. Instead, the node continues from where the previous node stopped processing the messages.
```
kbostrom, line 66: Do we need the second sentence; is this not implied by the first?

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
```diff
@@ -9,88 +9,90 @@ You can use Anypoint Connector for Salesforce (Salesforce Connector) as an inbou
...
+** If the cluster stops, the data from the object store is lost. When the cluster restarts, the configuration at the source level determines whether the connector reprocesses the messages.
 
-Having multiple sources listening to the same events and having the *Resume from the Last Replay Id* option enabled leads to data corruption and potential loss of events.
++
+To persist data so that the app continues processing events when restarted, create a persistent object store, as described in xref:mule-runtime::creating-and-managing-a-cluster-manually.adoc[Create and Manage a Cluster Manually].
+* If you use multiple workers when running the app in CloudHub, all of the workers receive and process the events.
```
kbostrom, line 71: Repeats Line 54?

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
```diff
@@ -9,88 +9,90 @@ You can use Anypoint Connector for Salesforce (Salesforce Connector) as an inbou
...
 
-=== Streaming Channel with No Listener
+== Multiple Flows That Use the Same Source
 
-Users can push events to a streaming channel, even if the channel does not have a listener for reading published events. Once a listener is started for the channel, Salesforce Streaming API pushes as many messages it can to the listener, based on the maximum daily limit.
+Avoid having two or more flows use a source that listens to events on the same channel or topic, because the events will be consumed twice from your user quota. If you have a use case that requires processing the same events in two or more different flows, consider implementing that logic in a single flow instead.
```
kbostrom: Just a couple alternatives for your consideration:
```suggestion
If you have two or more flows use a source that listens to events on the same channel or topic, the events are consumed twice from your user quota. To avoid this, consider implementing that logic in a single flow instead.
```
or
```suggestion
To avoid events being consumed twice when you have two or more flows use a source that listens to events on the same channel or topic, consider implementing that logic in a single flow instead.
```

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
```diff
@@ -9,88 +9,90 @@ You can use Anypoint Connector for Salesforce (Salesforce Connector) as an inbou
...
 
-For example, given that the maximum number of delivered event notifications within a 24-hour period for a free Salesfoce organization is 10,000, suppose you publish 15,000 events to that channel. When Salesforce Connector subscribes to that channel, the Streaming API attempts to push 10,000 events, thus consuming the daily quota. The API will then attempt to push the remaining 5,000 events the next day before pushing any new events.
+WARNING: Having multiple sources listening to the same events and having the *Resume from the Last Replay Id* option enabled can lead to data corruption and the potential loss of events.
```
kbostrom, line 78: BREAK HERE

kbostrom (dismissed):
Part 1 of review

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
kbostrom: RESUME

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
```diff
@@ -1,96 +1,97 @@
...
+Work with topic events by creating a topic, subscribing to a topic, and replaying topic messages.
 
 [[createtopic]]
-== Create a Topic to Receive Data from Salesforce
+=== Create a Topic to Receive Data from Salesforce
 
 When you create a topic, the connector creates a `PushTopic`, which is a special object in Salesforce that binds a name (in this case, the topic's name) and a Salesforce Object Query Language (SOQL) query together. After you create a topic, you can then subscribe to it by its name.
 
-You can use either the *Create* (`create`) or *Publish topic* (`publish-topic`) operation to create a topic. The following example uses the `publish-topic` operation to create a topic:
+You can use either the *Create* (`create`) operation or *Publish topic* (`publish-topic`) operation to create a topic. The following example uses the `publish-topic` operation to create a topic:
```
kbostrom, line 88:
```suggestion
You can use either the *Create* (`create`) operation or *Publish topic* (`publish-topic`) operation to create a topic. The following example uses `publish-topic` to create a topic:
```

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
```diff
@@ -103,32 +104,10 @@ insert pushTopic;
...
-    name="/u/Notifications"
-    description="General notifications"/>
-----
-
 [[topicsub]]
-== Subscribe to a Topic
+=== Subscribe to a Topic
 
-To subscribe to a topic, add the *Subscribe topic listener* (`subscribe-topic-listener`) or *Replay topic listener* (`replay-topic-listener`) as an input source for your flow. The input source acts as an inbound endpoint. Every time the subscription receives an event, the input source executes the rest of the flow in your Mule app.
+To subscribe to a topic, add the *Subscribe topic listener* (`subscribe-topic-listener`) or *Replay topic listener* (`replay-topic-listener`) as a source for your flow. The source acts as an inbound endpoint. Every time the subscription receives an event, the source executes the rest of the flow in your Mule app.
```
kbostrom:
```suggestion
To subscribe to a topic, add either the *Subscribe topic listener* (`subscribe-topic-listener`) or the *Replay topic listener* (`replay-topic-listener`) as a source for your flow. The source acts as an inbound endpoint. Every time the subscription receives an event, the source executes the rest of the flow in your Mule app.
```

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
```diff
@@ -142,58 +121,32 @@ In the following XML example, Mule prints a message to the log at the INFO level
...
-----
-
-The *Streaming channel* field of the *Subscribe channel listener* operation does not display change events that are available in the Salesforce environment. However, your connector can subscribe to a streaming channel to obtain this information. For example, to subscribe to the `All Change Events` channel, use `/data/ChangeEvents` as the channel name to which to subscribe.
-
-For more information, see https://developer.salesforce.com/docs/atlas.en-us.change_data_capture.meta/change_data_capture/cdc_subscribe_channels.htm[Subscription Channels] in the Salesforce Change Data Capture Developer Guide.
+You can subscribe to a topic that was not previously published in Salesforce. However, after the topic is published, you won't receive notifications for that topic unless you resubscribe to it.
 
 [[topicrep]]
-== Replay Messages From a Topic
+=== Replay Messages From a Topic
```
kbostrom:
```suggestion
=== Replay Messages from a Topic
```

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
```diff
@@ -142,58 +121,32 @@ In the following XML example, Mule prints a message to the log at the INFO level
...
-Subscriber receives new events that are broadcast after the client subscribes.
+Subscriber receives new events that are sent after the client subscribes.
 * `FROM_REPLAY_ID`
 +
 Subscriber receives all events after the specified event `replayId`.
 
 If you specify either the `ALL` replay option or `ONLY_NEW` replay option, the `replayId` value is ignored.
 
-The *Resume from the Last Replay Id* checkbox enables you to specify an automatic replay of stored events, based on the Replay ID of the last event processed by the connector. You can use this functionality when the connector stops listening, such as a during a server shutdown or dropped connection. If the stored Replay ID is outside the 24-hour retention period, the replay option determines what events to replay.
+The *Resume from the Last Replay Id* checkbox enables you to specify an automatic replay of stored events based on the Replay ID of the last event processed by the connector. You can use this functionality when the connector stops listening, such as a during a server shutdown or dropped connection. If the stored Replay ID is outside the 24-hour retention period, the replay option determines what events to replay.
```
kbostrom:
```suggestion
The *Resume from the Last Replay Id* checkbox enables you to specify an automatic replay of stored events based on the Replay ID of the last event processed by the connector. You can use this functionality when the connector stops listening, such as a during a server shutdown or dropped connection. If the stored Replay ID is outside the 24-hour retention period, the replay option determines which events to replay.
```

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
```diff
@@ -205,10 +158,55 @@ In the following XML example, the *Replay topic listener* operation acts like an
...
+[[createchannel]]
+=== Create a Streaming Channel
+
+To create a streaming channel, you must have the proper Salesforce Streaming API permissions enabled in your organization.
+
+Follow these steps to create a streaming channel:
+
+. Log in to your Salesforce Developer Edition organization.
+. Under *All Tabs (+)*, select *Streaming Channels*.
+. On the *Streaming Channels* tab, select *New* to create a new streaming channel.
```
kbostrom:
```suggestion
. On the *Streaming Channels* tab, select *New*.
```

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
```diff
@@ -205,10 +158,55 @@ In the following XML example, the *Replay topic listener* operation acts like an
...
+</flow>
+----
+
+The *Streaming channel* field of the *Subscribe channel listener* operation does not display change events that are available in the Salesforce environment. However, your connector can subscribe to a streaming channel to obtain this information. For example, to subscribe to the `All Change Events` channel, use `/data/ChangeEvents` as the channel name to which to subscribe.
+
+For more information, see https://developer.salesforce.com/docs/atlas.en-us.change_data_capture.meta/change_data_capture/cdc_subscribe_channels.htm[Subscription Channels] in the Salesforce Change Data Capture Developer Guide.
+
 [[channelrep]]
-== Replay Messages From a Streaming Channel
+=== Replay Messages From a Streaming Channel
```
kbostrom:
```suggestion
=== Replay Messages from a Streaming Channel
```

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
```diff
@@ -222,14 +220,28 @@ A streaming channel can replay notifications. The *Replay channel listener* inpu
 
 If you specify either the `ALL` replay option or the `ONLY_NEW` replay option, the `replayId` value is ignored.
 
-The *Resume from the Last Replay Id* checkbox enables you to specify an automatic replay of stored events, based on the Replay ID of the last event processed by the connector. You can use this functionality when the connector stops listening, such as a during a server shutdown or dropped connection. If the stored Replay ID is outside the 24-hour retention period, the replay option determines what events to replay.
+The *Resume from the Last Replay Id* checkbox enables you to specify an automatic replay of stored events based on the Replay ID of the last event processed by the connector. You can use this functionality when the connector stops listening, such as a during a server shutdown or dropped connection. If the stored Replay ID is outside the 24-hour retention period, the replay option determines what events to replay.
```
kbostrom: ditto "what" > "which"

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
```diff
@@ -222,14 +220,28 @@ A streaming channel can replay notifications. The *Replay channel listener* inpu
...
+
+To support the *Resume from Last Replay ID* feature, the connector uses a persistent object store to keep details regarding the processed messages. This feature reduce the possibility of message loss and avoids processing duplicate messages. For more information about how the object store is used, see <<objectstoreusage,Object Store Usage in Events>>.
```
kbostrom:
```suggestion
To support the *Resume from Last Replay ID* feature, the connector uses a persistent object store to keep details regarding the processed messages. This feature reduces the possibility of message loss and avoids processing duplicate messages. For more information about how the object store is used, see <<objectstoreusage,Object Store Usage in Events>>.
```

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
```diff
@@ -222,14 +220,28 @@ A streaming channel can replay notifications. The *Replay channel listener* inpu
...
+
+=== Push Events to a Streaming Channel with No Listener
+
+Users can push events to a streaming channel, even if the channel does not have a listener for reading published events. Once a listener is started for the channel, Salesforce Streaming API pushes as many messages it can to the listener, based on the maximum daily limit.
```
kbostrom:
```suggestion
Users can push events to a streaming channel, even if the channel does not have a listener for reading published events. After a listener is started for the channel, Salesforce Streaming API pushes as many messages as it can to the listener, based on the maximum daily limit.
```

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
```diff
@@ -222,14 +220,28 @@ A streaming channel can replay notifications. The *Replay channel listener* inpu
...
 
-To support the *Resume from Last Replay ID* feature, the connector uses a persistent object store to keep different details regarding the processed messages. This feature is enhanced in Salesforce Connector 10.x to reduce message loss probability and to avoid processing duplicate messages. For more details about how Object Store is used, check the <<objectstoreusage,Object Store Usage>> section.
+For example, given that the maximum number of delivered event notifications within a 24-hour period for a free Salesforce organization is 10,000, suppose you publish 15,000 events to that channel. When Salesforce Connector subscribes to that channel, Streaming API attempts to push 10,000 events, thus consuming the daily quota. The API will then attempt to push the remaining 5,000 events the next day before pushing any new events.
```
kbostrom:
```suggestion
For example, given that the maximum number of delivered event notifications within a 24-hour period for a free Salesforce organization is 10,000, suppose you publish 15,000 events to that channel. When Salesforce Connector subscribes to that channel, Streaming API attempts to push 10,000 events, thus consuming the daily quota. The API then attempts to push the remaining 5,000 events the next day, before pushing any new events.
```

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
```diff
@@ -222,14 +220,28 @@ A streaming channel can replay notifications. The *Replay channel listener* inpu
...
+
+In this scenario, the connector streams the events one-by-one into the Mule app. If the app takes too long to process a message, Streaming API might instruct the connector to reconnect. If this happens, Streaming API drops all of the unprocessed messages.
+
+You can avoid this situation by implementing a reliability pattern, as described in xref:mule-runtime::reliability-patterns.adoc[Reliability Patterns].
```
kbostrom: Consider making this a single para.

kbostrom (dismissed):
Second half of review

salesforce/10.10/modules/ROOT/pages/salesforce-connector-processing-events.adoc
stefangretcan: There are 2 things: the order in which the events are received from Salesforce. And second: the order in which the events are processed by the Mule flow that is right next to the source. The evets come from Salesforce in incremental order. For example: 2 3 4 5 7 8 9 11. Now, the source might process them like this: 11 3 4 5 2 7 8 9.
Let's suppose event with replayId = 3 fails. Then we save as lowest failed replay id the number 3. Then, 4 and 5 are processed successfully. But event with replayId = 2 fails. Now the lowest failed replay id that we save is: 2. 
If the source is now restarted and it has the "Resume from last replay id" option selected, then it will start from 2.

stefangretcan: The actual replay id is saved. I do not think the user needs to know what we save internally. Only thing they need to know is that the restart logic of a source that has the "Resume from last replay id" option selected is:
1) If there are one or more failed events, the source will start with the lowest replay id that failed;
2) If there are no failed events, the source will start from the last successfully processed event id.

stefangretcan (dismissed):

dhirsch1:
Closed this PR to avoid merge conflicts. The topic in this PR was moved to https://github.com/mulesoft/docs-connectors/pull/1484.

PR #1463, opened 2021-05-13
DOCS-10688-1: Amazon Lambda Examples

Hazarth:
added an usage example for lambda


amazon-lambda/1.0/modules/ROOT/pages/amazon-lambda-connector-examples.adoc
```diff
@@ -0,0 +1,88 @@
...
+{
+    "first_name":"Tyler",
+    "last_name":"Durden"
+}
+----
+
+The Mule flow responds with a code 200 message:
+----
+{
+    "message": "Hello <first_name> <last_name>!"
```
Hazarth, line 79: @valkyrie69 
you changed the response to `Hello <first_name> <last_name>!`
but the payload is still
```
{
    "first_name":"Tyler",
    "last_name":"Durden"
}
```

shouldn't we also change the payload to 
```
{
    "first_name":"<first_name>",
    "last_name":"<last_name>"
}
```

so the result matches the payload? I think it might end up being a bit confusing if they don't match up. This is also why I originally chose a real example name, so it's clear that A leads to B exactly, which should match the python code.

PR #1466, opened 2021-05-13
Salesforce Examples topic revamp

dhirsch1:
…ple into the Studio editor


sathya0:
ported this info here https://github.com/mulesoft/docs-connectors/pull/2124

PR #1506, opened 2021-06-09
Added the batch topic back to the X12 Connector docs

dhirsch1:


dhirsch1:
Starting again - just need to fix the nav

PR #1529, opened 2021-06-25
DOCS-11360 LDAP 3.4

sathya0:


ldap/3.4/modules/ROOT/pages/ldap-connector-reference.adoc
```diff
@@ -0,0 +1,1105 @@
...
+| Name | Type | Description | Default Value | Required
+| Principal DN a| String |  |  | x
+| Password a| String |  |  |
+| Authentication a| String |  |  |
+| URL a| String |  |  | x
+| Reconnection a| <<Reconnection>> |  When the application is deployed, a connectivity test is performed on all connectors. If set to `true`, deployment will fail if the test doesn't pass after exhausting the associated reconnection strategy. |  |
+| Type a| Enumeration, one of:
+
+** JNDI |  The implementation of the connection to use. |  JNDI |
+| Schema Enabled a| Boolean |  If set to `true`, the LDAP connector will use the LDAP schema (works only for LDAP v3) to define the structure of the LDAP entry or map. Must be set to `true` to use DataSense as it will affect the implementing class of `org.mule.module.ldap.api.LDAPEntry` attributes. |  false |
```
valkyrie69:
```suggestion
| Schema Enabled a| Boolean |  If set to `true`, the LDAP connector will use the LDAP schema (works only for LDAP v3) to define the structure of the LDAP entry or map. This parameter must be set to `true` to use DataSense, as it affects the implementing class of `org.mule.module.ldap.api.LDAPEntry` attributes. |  false |
```

ldap/3.4/modules/ROOT/pages/ldap-connector-reference.adoc
```diff
@@ -0,0 +1,1105 @@
...
+| Referral a| Enumeration, one of:
+
+** IGNORE
+** THROW
+** FOLLOW |  A constant that consists of the name of the environment property to specify how referrals encountered by the service provider must be processed (FOLLOW, IGNORE, THROW). | IGNORE |
+| Extended Configuration a| Object |  A map instance that consists of extended configuration attributes used in the context environment. When working with TLS connections, ensure that the native LDAP pooling functionality is turned off. For example, if using JNDI do not use attributes such as `com.sun.jndi.ldap.connect.pool=true` as it will cause problems when using TLS. |  |
```
valkyrie69:
```suggestion
| Extended Configuration a| Object |  A map instance that consists of extended configuration attributes used in the context environment. When working with TLS connections, ensure that the native LDAP pooling functionality is turned off. For example, if using JNDI, do not use attributes such as `com.sun.jndi.ldap.connect.pool=true`, as it will cause problems when using TLS. |  |
```

ldap/3.4/modules/ROOT/pages/ldap-connector-reference.adoc
```diff
@@ -0,0 +1,1105 @@
...
+|===
+
+== SSL Configuration
+
+=== Parameters
+
+[%header%autowidth.spread]
+|===
+| Name | Type | Description | Default Value | Required
+| Principal DN a| String |  |  | x
```
valkyrie69: All of these fields should have descriptions. This is an issue with the java files. We should open a ticket against the connector to add descriptions for these fields. You can add them manually if you have time for this release but then we would have to get a technical review. Like, for Principal DN, does it accept either the name of the domain OR the IP address? Or just the name? 

Principal DN: Name or IP address of the domain used by your LDAP implementation.

valkyrie69: Actually, I don't even know for sure that DN is referring to domain name.

valkyrie69 (dismissed):

valkyrie69 (approved):

PR #1501, opened 2021-06-07
DOCS-11124 Salesforce CDP Connector 1.0

valkyrie69:

PR #1541, opened 2021-06-30
Updated index, studio, config and xml pages, renamed to Salesforce CDP from Salesforce Data Cloud

adatta0728:
Renamed the folders, files and content to salesforce-cdp from salesforce-data-cloud
Added content for the pages below - 
 - Index Page
 - Studio Page
 - XML and Maven Support page
 - Additional Configuration Topics page


valkyrie69:
The PR being used for this content is #1501 These updates can be found in that PR.

PR #1546, opened 2021-07-01
Data cloud examples

DonaldBough:
All the streaming and bulk api example flows, branched for the original Data Cloud connector branch. All references to Data Cloud should already be removed, but feel free to double-check.


valkyrie69:
These updates were added to PR #1562

PR #1552, opened 2021-07-14
DOCS-11447 Box 5.1.1

sathya0:


box/5.1/modules/ROOT/pages/box-connector-reference.adoc
```diff
@@ -8836,7 +8836,7 @@ Retrieves folder lock details for a given folder. This operation makes an HTTP G
 `<box:get-folders-by-folder-id>`
 
 
-Retrieves details for a folder, including the first 100 entries in the folder. To fetch more items within the folder, use the [Get items in a folder](#get-folders-id-items) endpoint. This operation makes an HTTP GET request to the /folders/{folder_id} endpoint.
+Retrieves details for a folder, including the first 100 entries in the folder. To fetch more items within the folder, use the [Get items in a folder](#get-folders-id-items) endpoint. This operation makes an HTTP GET request to the /folders/{folder_id} endpoint. This operation is automatically paginated with offset pagination.
```
dhirsch1: IMHO, the operation is not paginated - the operation results are. 

I suggest: The operation results are automatically paginated with offset pagination.
or
The operation automatically paginates the results using offset pagination. (I am assuming here that the operation is doing the pagination but if you're not sure, use the first suggestion instead.)

hreso110100: I think that original description is more precise as connector's operation is performing pagination when doing request

adam-lipka: To me, all three options (the original and two Diane's) sound good, are all valid, and really there is no huge difference. Whatever sounds best to you.

dhirsch1 (approved):

PR #1574, opened 2021-07-23
Docs 11369 ftp connector about page

luanamulesoft:


kevintroller (dismissed):
LGTM

ftp/1.5/modules/ROOT/pages/index.adoc
```diff
@@ -3,68 +3,35 @@
 
 Support Category: https://www.mulesoft.com/legal/versioning-back-support-policy#anypoint-connectors[Select]
 
-FTP Connector v1.5
-
-Anypoint Connector for FTP (FTP Connector) provides access to files and folders on an FTP server. Its main features include:
+Anypoint Connector for FTP (FTP Connector) provides access to files and folders on an FTP server. FTP connector operations manage file transfers over the FTP (File Transfer Protocol) protocol. The connector's main features include:
```
kbostrom:
```suggestion
Anypoint Connector for FTP (FTP Connector) provides access to files and folders on an FTP server. FTP connector operations manage file transfers by using File Transfer Protocol (FTP). The connector's main features include:
```

ftp/1.5/modules/ROOT/pages/index.adoc
```diff
@@ -3,68 +3,35 @@
...
-  <artifactId>mule-ftp-connector</artifactId>
-  <version>RELEASE</version>
-  <classifier>mule-plugin</classifier>
-</dependency>
-----
-
-Mule converts RELEASE to the latest version. To specify a version, view
-https://www.mulesoft.com/exchange/org.mule.connectors/mule-ftp-connector/[FTP Connector] in
-Anypoint Exchange and click *Dependency Snippets*.
+* A design that is consistent with the File and FTP connectors.
```
kbostrom: Since "File" is init. capped, it is a connector name, and connector names always capitalize "Connector":
```suggestion
* A design that is consistent with the File and FTP Connectors.
```

ftp/1.5/modules/ROOT/pages/index.adoc
```diff
@@ -3,68 +3,35 @@
...
++
+The same set of operations is available on these connectors, and they behave similarly.
```
kbostrom:
```suggestion
The same set of operations is available using these two connectors, and they behave similarly.
```

ftp/1.5/modules/ROOT/pages/index.adoc
```diff
@@ -3,68 +3,35 @@
...
 
-[[connection_settings]]
-== Connect to an FTP Server
+=== Representation of File Attributes
 
-You can set up a connection through an `ftp:config` element, for example:
+Mule 4 provides a message structure of a payload and attributes in which the data moves around, making that payload a file content. The attributes provide an object that contains metadata on the payload and information such as the file’s name, path, size, and timestamps. For FTP Connector and SFTP Connector, these attributes are known as  _file attributes_.
```
kbostrom, line 18: I'm not sure about that first sentence: is the fact that the data moves around make the payload file content? Isn't a payload always file content.

kbostrom: I also wonder if we really need to point out to this audience what a "file attribute" is. I feel like we are defining terms with which the audience is already familiar: payload, attributes, file. But perhaps I am missing something?

kbostrom: My suggestion, if it matches your intention, would be something like this:
```suggestion
In Mule 4, messages comprise _file content_ (a payload) and _file attributes_ (metadata about the payload, such the file’s name, path, size, and timestamps.
```

kbostrom: ...but is this any different from previous Mule versions?

ftp/1.5/modules/ROOT/pages/index.adoc
```diff
@@ -3,68 +3,35 @@
...
-<ftp:config name="ftp">
-  <ftp:connection
-    username="anonymous"
-    password="password"
-    host="localhost"
-    port="${ftpPort}"
-    workingDir="${workingDir}"/>
-</ftp:config>
-----
+The Mule message contains the concepts of encoding and MIME type. These describe the format in which a payload is represented. The file attributes describe a file’s encoding and MIME type.
```
kbostrom: These sentences repeat "encoding and MIME type," which is probably unnecessarily wordy.
How about tacking this info onto the previous?
```suggestion
In Mule 4, messages comprise _file content_ (a payload) and _file attributes_ (metadata about the payload, such the file’s name, path, size, and timestamps. The file attributes also describe the format in which a payload is represented by including a file’s encoding and MIME type.
```

kbostrom (dismissed):
CE, CU

ftp/1.5/modules/ROOT/pages/index.adoc
luanamulesoft: Hey Kim!  As these FTP Connector pages are based on the SFTP Connector pages (the content is very similar, so I based my content on them), I had to check on these content questions, and I learned a lot 😃 
About the payload "moving around", I found that payload content changes (message processors in the flow make these changes) as it travels through a flow. The Payload can contain the response to an HTTP request, records retrieved from a database, or the content of a file (in the case of FTP and SFTP connectors). So, I think payload content does change because it moves around.
As per the attributes, I found that the specific attributes of a message depend on the connector (such as HTTP, FTP, File) associated with the message, in the context of the Mule message structure. So, I think that is why it is being defined.

kevintroller (approved):

PR #1592, opened 2021-08-12
DOCS-11777: Addressed issue in docs

hannanelson:


anypoint-mq/3.x/modules/ROOT/pages/anypoint-mq-ack.adoc
```diff
@@ -51,7 +51,14 @@ By default, the connector honors the time defined at the queue level (*Message L
...
 
 [IMPORTANT]
-When using the `prefetch` mode in the Anypoint MQ Subscriber source, the time that the message remains in the local buffer doesn’t count as part of the timeout. The `acknowledgementTimeout` timer begins when the message is dispatched to the flow, not when it is taken from the queue. The default TTL is two minutes. If you need a different TTL, you must define it explicitly in the `acknowledgementTimeout` parameter of the subscriber.
+----
+When using the `prefetch` mode in the Anypoint MQ Subscriber source, the time that the message remains in the local buffer doesn’t count as part of the timeout. The `acknowledgementTimeout` timer begins when the message is consumed. The default TTL is two minutes. If you need a different TTL, you must define it explicitly in the `acknowledgementTimeout` parameter of the subscriber.
+
+If the time that the message remains in the buffer reaches 80% of the timeout, the expiration time is extended.
```
bahui80: Taking into account this last sentence, shouldn't then the previous one be deleted?

bahui80: What about saying: `when using the prefetch mode in the Anypoint MQ Subscriber source, ...` and saying something like the timeout will try to be preserved. Like trying to make notice that this is not always true. 
That the timeout will try to be honoured when possible, and that is when 80% of time was reached in the internal buffer or it actually expired.

anypoint-mq/3.x/modules/ROOT/pages/anypoint-mq-listener.adoc
```diff
@@ -27,7 +27,10 @@ When a consumer takes a message from the queue and stores it in the local buffer
 +
 The smaller buffer restricts app-level throughput and avoids blocking competing consumers.
 
-When using the `prefetch` mode, the `acknowledgementTimeout` timer begins from the moment the message is dispatched to the flow, not when it is taken from the queue. The time the message remains in the local buffer doesn't count as part of the timeout.
+When using the `prefetch` mode, the `acknowledgementTimeout` timer begins from the moment the message is taken from the queue. The time the message remains in the local buffer doesn't count as part of the timeout.
```
bahui80: Same here, it actually does, the message in the local buffer counts as part of the timeout except when we reach the previously mentioned condition

anypoint-mq/3.x/modules/ROOT/pages/anypoint-mq-ack.adoc
hannanelson: See the change. I'm not sure we want to say the timeout isn't honored if it expired (do we?) Isn't that expected behavior?

anypoint-mq/3.x/modules/ROOT/pages/anypoint-mq-listener.adoc
```diff
@@ -27,7 +27,11 @@ When a consumer takes a message from the queue and stores it in the local buffer
 +
 The smaller buffer restricts app-level throughput and avoids blocking competing consumers.
 
-When using the `prefetch` mode, the `acknowledgementTimeout` timer begins from the moment the message is dispatched to the flow, not when it is taken from the queue. The time the message remains in the local buffer doesn't count as part of the timeout.
+When using the `prefetch` mode, the `acknowledgementTimeout` timer begins when the message is taken from the queue. 
+Anypoint MQ attempts to honor the expiration time specified by `acknowledgementTimeout`.
+However, Anypoint MQ extends the expiration timeout if the time that the message remains in the buffer reaches 80% of the timeout value.
+
+Don't use `prefetch` mode for applications with long processing times or that require messages to be processed only once.
```
bahui80: I would remove the part of `that require messages to be processed only once.` I know it's specific to what this customer saw but its too generic and in most cases this is not true. 
I would just leave the first part and probably do:
`It is not recommended to use prefetch mode for applications with long processing times and max concurrency limitation since message will remain in buffer for some time` -> change to something nicer jajaj. I am not good at writing this stuff

schinthakindims (approved):
LGTM

bahui80 (approved):
LGTM!

valkyrie69 (approved):

PR #1671, opened 2021-10-21
DOCS-12427 NetSuite 11.5

sathya0:


netsuite/11.5/modules/ROOT/pages/netsuite-reference.adoc
```diff
@@ -0,0 +1,2506 @@
...
+** DAYS |  The time unit for Connection Timeout. |  MILLISECONDS |
+| Base Netsuite Address a| String |  The base address to use to connect to NetSuite. The default value is the main NetSuite site. This is used to automatically obtain the endpoint for the service based on the credentials configured, and will need no change. |  https://webservices.netsuite.com |
+| Wsdl version a| Enumeration, one of:
+
+** V2020_2
+** V2021_1 a|  Version of WSDL that the connector uses to communicate with NetSuite:
+
+* If you do not select a value, the connector works with SuiteTalk version `2020pass:[_]2` and existing namespaces in the Mule app must not be modified.
+
+* If you select a WSDL version, remove the SuiteTalk version from all namespaces in the Mule app. The following example shows an original DataWeave script with the SuiteTalk version and a new DataWeave script with the SuiteTalk version removed.
```
dhirsch1: I suggest removing the example intro from the bullet and putting it on its own line. You are right that it just applies to the second bullet, but the alternative is to use plus signs to indent it under the bullet and because the table column may not be wide enough, that may not work. You could try it though.

dhirsch1 (approved):
Approved as long as the example is not chopped in the local build.

PR #1711, opened 2021-11-19
SE-23558: update docs for Google Pub/Sub PatchOperation 

IvoUlbricht:


google-pubsub/1.0/modules/ROOT/pages/google-pubsub-connector-reference.adoc
```diff
@@ -298,8 +298,8 @@ Updates an existing snapshot.
 |===
 | Name | Type | Description | Default Value | Required
 | Configuration | String | Name of the configuration to use. | | x
-| Snapshot a| Any |  Snapshot object with updated fields. |  #[payload] |
-| Update Mask a| String |  Indicates which fields in the provided snapshot to update. This is a comma-separated list of fully qualified names of fields. |  | x
+| Snapshot a| <<Snapshot>> |  Snapshot object with updated fields (labels or expireTime). |  #[payload] |
```
dhirsch1, line 301: Not sure what labels and expireTime are. Are they fields? Can you write them as they are in Studio (preferred way to do this for new text)? The text says updated fields, and these two are listed with an or. Is there just one updated field?

dhirsch1 (approved):

PR #1728, opened 2021-12-11
Update kafka-connector-studio.adoc

DerejeKitaw:
Consumer connection type updated to match 4.6 kafka connector connector


sathya0:
Hi @DerejeKitaw! Thanks for the contribution. It looks like all of the connection types use the Bootstrap Server URLs and Bean reference, not just the Consumer Plaintext Connection. The reason why the Consumer Plaintext Connection doesn't have its own row is because it doesn't have any extra fields. Let me know if you have any other comments or questions. If this is resolved, I will close the PR. Thank you.

PR #1729, opened 2021-12-13
DOCS-12657 Shopify 1.1

sathya0:


shopify/1.1/modules/ROOT/pages/index.adoc
```diff
@@ -0,0 +1,52 @@
...
+
+== Connection Types
+
+Shopify Connector uses the following authentication methods:
+
+* Basic Auth +
+Uses a username and password for authentication.
+
+* OAuth2 +
+Delegates user authentication to the service hosting the user account.
```
tasmoore, line 41: Should this be "...user's account" instead of "user account?"

tasmoore (approved):
Just a comment on the new sources. Wondered if the Output table should have a content row below the Type/Any header.

PR #1768, opened 2022-01-24
DOCS-13083-verified

kevintroller:
Squash merge to verify commits


dhirsch1 (approved):
Just a suggestion for the future - change "key" to "major" or something similar since "key" in this context can also mean "differences in keys."

PR #1777, opened 2022-01-26
Exceptions needs to be thrown and no ignored

ca-stefan-cordes:
as they cause data inconsistence when continue the flow.
See https://www.mulesoft.org/jira/browse/DBCON-199

### DIFF
--- a/db/1.11/modules/ROOT/pages/database-connector-troubleshooting.adoc
+++ b/db/1.11/modules/ROOT/pages/database-connector-troubleshooting.adoc
@@ -79,59 +79,7 @@ The following example shows how to use pooling profiles with the *Stored procedu
 	</flow>
 ----
 
-== Understand Select and Stored Procedure Operations Warning Exceptions
 
-When the Database Connector *Select* and *Stored procedure* operations execute correctly but do not return results, Mule runtime engine (Mule) returns a `WARN` exception message instead of throwing the exception.
-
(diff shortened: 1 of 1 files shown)


kevintroller:
Closing as the requested change does not correlate with the reported JIRA issue (which has already been fixed).

ca-stefan-cordes:
https://www.mulesoft.org/jira/browse/DBCON-199 is still open.

kevintroller:
Hi @ca-stefan-cordes, our Docs team reviewed this case with the dev team and it does not correlate with the described issue in the JIRA and subsequent comments. Could you please provide additional information in the JIRA why you consider that the section called "Understand Select and Stored Procedure Operations Warning Exceptions" should be removed from the Database Connector documentation? Thank you!

ca-stefan-cordes:
Hi @kevintroller ,
https://www.mulesoft.org/jira/browse/DBCON-199 is requesting to throw an exception in case "ResultSet not consumable" instead of silently "warn" about lost data (lost rows) and continue processing.

Behavior just to "warn" is based on documentation of "database-connector-troubleshooting.adoc" which states:
* some drivers may not work correctly.
* so all applications will not fail but warn.

Silently warning of programming errors (consuming streamable resultset out of transaction) will create inconsistent data and it takes decades to find the root cause for it.

So only solution from my side is to throw an exception (and so documentation section is obsolete which is this pull request).

In case you fear problems you may update the documentation to "how-to" which describes 
which properties may be set for "but other drivers do not" to prevent exceptions.

PR #1818, opened 2022-02-22
W-10729984 S/4HANA SOAP 2.3

wwarner-salesforce:
W-10729984 S/4HANA SOAP 2.3

PR #1826, opened 2022-03-01
W-10743083 S/4HANA SOAP non-repeatable streams

wwarner-salesforce:
W-10743083 S/4HANA SOAP non-repeatable streams


sathya0 (approved):

wwarner-salesforce:
This was in the 2.2 docs and removed in https://github.com/mulesoft/docs-connectors/pull/1818 . Devs explained that it was still valid for 2.3, even though it was not in the generated reference file.

PR #1839, opened 2022-03-18
Email Connector - Correct the port number to match the associated diagram - W-10587723

dhirsch1:


sathya0 (dismissed):
Might have to change the port number in the XML as well

sathya0 (approved):

PR #1852, opened 2022-03-24
W-10824433 SAP manage transactions example

sathya0:


sap/5.5/modules/ROOT/pages/sap-connector-examples.adoc
```diff
@@ -690,6 +698,499 @@ output application/java
...
+----
+. From the *Mule Palette* view, select *SAP* and drag another *Synchronous Remote Function Call* operation next to the *Transform Message* component.
+. Create a global element named `SAP_Config` and specify the connection information.
+. Click *Test Connection* to confirm that Mule runtime engine can connect with the SAP instance.
+. Enter `"BAPI_BANK_GETDETAIL"` for the *Function Name* and `payload` for the *Content*.
+. In the *Mule Palette* view, select *Core* and drag a *Logger* component next to *Synchronous Remote Function Call*.
+
+=== Configure the Second Flow
+
+The second flow creates a sales order from the data. To configure this flow:
```
jorgegarciamule: Here we are adding a Try around the BAPI call to create a transaction, if not, the change does not impact on SAP.

sap/5.5/modules/ROOT/pages/sap-connector-examples.adoc
```diff
@@ -690,6 +698,499 @@ output application/java
...
+image::managetransactions-flow1.png[Studio flow for the first flow of Managing Transactions]
+
+image::managetransactions-flow2.png[Studio flow for the second flow of Managing Transactions]
+
+image::managetransactions-flow3.png[Studio flow for the third flow of Managing Transactions]
+
+
+=== Configure the First Flow
+
+The first flow retrieves a list of available banks in the United States, filters the first result, and fetches information about the bank. To configure this flow:
```
jorgegarciamule: In this scenario, we are not working with transactions as it's not necessary for queries to SAP.

sap/5.5/modules/ROOT/pages/sap-connector-examples.adoc
```diff
@@ -690,6 +698,499 @@ output application/java
...
+. Click *Test Connection* to confirm that Mule runtime engine can connect with the SAP instance.
+. Enter `BAPI_SALESORDER_CREATEFROMDAT2` for the *Function Name* and `payload` for the *Content*.
+. From the *Mule Palette* view, select *Core* and drag *On Error Continue* under *Synchronous Remote Function Call*.
+. In the *Mule Palette* view, select *Core* and drag a *Logger* component into *On Error Continue*.
+. In the *Mule Palette* view, select *Core* and drag another *Logger* component next to *Synchronous Remote Function Call*.
+. Enter `payload.BAPI_SALESORDER_CREATEFROMDAT2.export.SALESDOCUMENT` for the *Message*.
+
+=== Configure the Third Flow
+
+The third flow creates a transaction ID and gets a list. To configure this flow:
```
jorgegarciamule: Async calls do not expect a response and require a different way of managing the transaction.  That's why we have specific operations for them. 
About the example, maybe doing a query may not be a useful scenario for an async. It would make more sense something that triggers a process or creates something on SAP.

jorgegarciamule (commented):
I added some comments that can be added to each scenario which tell a bit of the "why". 3rd scenario makes sense technically but not in real life.

jorgegarciamule (approved):

tasmoore (approved):

PR #1860, opened 2022-04-05
W-10950264: Added errors descriptions

acurci-at-mulesoft:
# Understand Common Throws

Here is a list of common throw messages and how to interpret them.

## FILE:CONNECTIVITY

The connection is no longer valid. It is disposed of, but you can apply a reconnection policy if defined in the application setup.

## FILE:ILLEGAL_PATH

If the *createParentDirectories* property is set to false and the target path doesn’t exist, and neither does its parent,
then a FILE:ILLEGAL_PATH error is thrown.

## FILE:FILE_ALREADY_EXISTS

If the *overwrite* property is set to false and the target file already exists, then a FILE:FILE_ALREADY_EXISTS error is thrown.

## FILE:RETRY_EXHAUSTED

The maximum number of retries for the operation is reached.

## FILE:ACCESS_DENIED

A File Connector operation attempted to access a file or a path and performed an action for which the user has not been granted permission.

## FILE:FILE_LOCK

An operation made an unsuccessful attempt to acquire a lock over a file.

## FILE:ILLEGAL_CONTENT

The Write operation received invalid content to write.


file/1.3/modules/ROOT/pages/file-troubleshooting.adoc
```diff
@@ -33,6 +33,39 @@ To resolve this error:
...
+
+Here is a list of common throw messages and how to interpret them:
+
+* FILE:CONNECTIVITY
+
+The connection to the file server is no longer valid. Apply a reconnection policy if defined in the application setup.
+
+* FILE:ILLEGAL_PATH
+
+This error occurs if the *createParentDirectories* property is set to false and the target path doesn’t exist nor does its parent.
```
kbostrom: Parallelism in list item wording:
```suggestion
The *createParentDirectories* property is set to `false` and neither the target path nor its parent exists.
```

file/1.3/modules/ROOT/pages/file-troubleshooting.adoc
```diff
@@ -33,6 +33,39 @@ To resolve this error:
...
+
+* FILE:FILE_ALREADY_EXISTS
+
+This error occurs when the target file already exists and the *overwrite* property is set to false.
```
kbostrom: ditto:
```suggestion
The target file already exists and the *overwrite* property is set to `false`.
```

file/1.3/modules/ROOT/pages/file-troubleshooting.adoc
```diff
@@ -33,6 +33,39 @@ To resolve this error:
...
+
+A File Connector operation attempted to access a file or a path and performed an action for which the user does not have permission.
+
+* FILE:FILE_LOCK
+
+An attempt to lock a file was unsuccessful.
+
+* FILE:ILLEGAL_CONTENT
+
+The *Write* operation received invalid content to write.
```
kbostrom, line 66: Does this mean the same thing?
```suggestion
The *Write* operation was attempted using invalid content.
```
I think this might be one of those cases in which passive voice makes the meaning more clear.

kbostrom (dismissed):
CE

dhirsch1 (approved):

PR #1869, opened 2022-04-08
W-10967201 - update pollingTime

hannanelson:


yunmei-zhang:
Hi @hannanelson, for the first change against the polling time in the table, the supported value should be from 0 to 20 000 milliseconds.
For the second change, kindly refer to GUS W-10967879. Better to leave the "Default" setting of polling time which is empty value and add another scenario for the polling time when it's out of supported range.

hannanelson:
> Hi @hannanelson, for the first change against the polling time in the table, the supported value should be from 0 to 20 000 milliseconds. For the second change, kindly refer to GUS [W-10967879](https://gus.my.salesforce.com/apex/ADM_WorkLocator?bugorworknumber=W-10967879). Better to leave the "Default" setting of polling time which is empty value and add another scenario for the polling time when it's out of supported range.

0 just defaults to 10 :|

anypoint-mq/3.x/modules/ROOT/pages/anypoint-mq-consume.adoc
```diff
@@ -52,17 +52,25 @@ For information about acknowledgment timeouts, see xref:anypoint-mq-ack.adoc#ack
 
 == Empty Queue Management
 
-When a queue has messages, the Consume operation returns immediately with one of the messages. When a queue is empty, the operation behavior depends on how you set the 'pollingTime' parameter:
+When a queue has messages, the Consume operation returns immediately with one of the messages. When a queue is empty, the operation behavior depends on the value of the `pollingTime` parameter.
+
+* No value
++
+By default, the Consume operation waits for 10000 milliseconds (10 seconds) for a new message to arrive when consuming from an empty queue. Consume fails if no message arrives in that timeframe.
```
GouthamiKondakindi: For consistency, we should mention what error Consume fails with(similar to the '1' to '20000' case):
If Anypoint MQ receives no message within this time window, the operation fails with an `ANYPOINT-MQ:TIMEOUT` error.

anypoint-mq/3.x/modules/ROOT/pages/anypoint-mq-consume.adoc
```diff
@@ -52,17 +52,25 @@ For information about acknowledgment timeouts, see xref:anypoint-mq-ack.adoc#ack
...
 
 * `0`
 +
-If you set the `pollingTime` parameter to `0`, the Consume operation returns immediately without any error and produces a `null` message as the result when the queue is empty.
+If the value of `pollingTime` is `0`, the Consume operation waits for the default of 10000 milliseconds (10 seconds).
```
GouthamiKondakindi, line 67: Should we keep this part? "produces a `null` message as the result when the queue is empty."
When pollingTime = 0, we don't throw an error like the other scenarios here.

anypoint-mq/3.x/modules/ROOT/pages/anypoint-mq-consume.adoc
hannanelson: @GouthamiKondakindi - it's the same type of error? ANYPOINT-MQ:TIMEOUT?

anypoint-mq/3.x/modules/ROOT/pages/anypoint-mq-consume.adoc
hannanelson: Yes, I don't know why I deleted it.

anypoint-mq/3.x/modules/ROOT/pages/anypoint-mq-consume.adoc
hannanelson: `ANYPOINT-MQ:CONSUMING` ?

GouthamiKondakindi: Yes, it's the same type of error "ANYPOINT-MQ:TIMEOUT"

hannanelson: OK, that wasn't clear from the error Yunmei added to the work item.

GouthamiKondakindi (approved):
LGTM

dukesphere (approved):

yunmei-zhang:
LGTM

PR #1879, opened 2022-04-13
W-10915923-vm-troubleshooting-kt

kevintroller:
W-10915923-vm-troubleshooting-kt


acurci-at-mulesoft:
@kevintroller I added another error to the list:
```
* VM:PUBLISH_CONSUMER_FLOW_ERROR
In a publish-consume operation, the flow that listens for the published message failed to process it.
```
Because I found it in the [code](https://github.com/mulesoft/mule-vm-connector/blob/2.0.0/src/main/java/org/mule/extensions/vm/internal/operations/PublishConsumeErrorTypeProvider.java#L29). 
I checked the [docs](https://docs.mulesoft.com/vm-connector/2.0/vm-reference#publishConsume) and it is not there. How do you get the information about the errors that are thrown?

kevintroller:
@acurci-at-mulesoft The error throws in the reference appear when the reference guide is auto generated from the Javadoc files of the connector. My best guess is that the next time that the reference guide is generated for VM Connector, it will bring this new error that you mentioned.

vm/2.0/modules/ROOT/pages/vm-connector-troubleshooting.adoc
```diff
@@ -0,0 +1,58 @@
...
+. Save your application changes.
+. Click the project name in *Package Explorer* and then click *Run* > *Run As* > *Mule Application*.
+
+
+== Understand Common Throws
+
+Here is a list of common throw messages and how to interpret them:
+
+* VM:EMPTY_QUEUE
+
```
sathya0, line 37: put + in between the bullet and description maybe

sathya0 (dismissed):

sathya0 (approved):

PR #1959, opened 2022-06-06
W-11256404 - fix connectors icons and exchange links

sanderson-sfdc:
ref: W-11256404

- In multiple antora files, added missing `page-exchange-group-id` and `page-exchange-asset-id` attributes which are needed to construct the icon and connector URLs
- add `page-icon-extension: png` for certain connectors that use png, instead of svg, for their icons. This attribute will work with the new UI bundle
- fix some IDs that were no longer valid, which resulted in broken links and images


cheungaryk (dismissed):

cheungaryk:
unverified commit...cannot merge

cheungaryk:
this PR should fix all but 2 images: generic and quartz (which does not exist anymore). Those will be handled separately - maybe we can use a default fallback image using one of the methods detailed in [this StackOverflow Q&A](https://stackoverflow.com/questions/980855/inputting-a-default-image-in-case-the-src-attribute-of-an-html-img-is-not-vali)

amazon-dynamodb/1.4/antora.yml
```diff
@@ -9,6 +9,8 @@ asciidoc:
     page-component-desc: Provides connectivity to the Amazon DynamoDB API.
     page-connector-type: Connector
     page-connector-level: Select
+    page-exchange-group-id: org.mule.module
```
sathya0:
```suggestion
    page-exchange-group-id: com.mulesoft.connectors
```

hdfs/6.0/antora.yml
```diff
@@ -9,4 +9,6 @@ asciidoc:
     page-component-desc: Works as a bidirectional gateway between Mule applications and Apache Hadoop Distributed File System (HDFS).
     page-connector-type: Connector
     page-connector-level: Select
+    page-exchange-group-id: org.mule.modules
```
sathya0:
```suggestion
    page-exchange-group-id: org.mule.connectors
```

sathya0 (approved):

PR #2029, opened 2022-07-28
W-11493682 veeva vault 1.4

sathya0:
# Writer's Quality Checklist

- [ ] Run spell checker
- [ ] Check for orphan files
  - Broken images
  - Dead links
  - Formatting issues, such as:
    - Messed up tables
    - Proper indentation
    - Correct header levels
  - Technical SME
  - Product Manager
  - Editor or peer reviewer


veevavault/1.4/modules/ROOT/pages/veevavault-connector-studio.adoc
```diff
@@ -0,0 +1,208 @@
...
+|Authorization Url | OAuth2 authorization URL, such as `\https://{your-instance}/as/authorization.oauth2` for PingFederate.
+|Access Token Url | OAuth2 token URL, such as `\https://{your-instance}/as/token.oauth2`.
+|Scopes | You can set up your OAuth2 scopes as necessary, but the important ones `openid` and `offline_access` are already pre-filled.
+|Resource Owner Id | Depending on your IdP configuration, you can leave it blank.
+|Listener Config | Create a new HTTP Listener or use an existing one if you have one, for example, `\http://localhost:8081/`.
+|Callback Path | Callback path of your application. If you set it to `callback`, your full callback path is `\http://localhost:8081/callback`. Do not forget to add your callback path to the IdP's redirect URLs.
+|Authorize Path | Authorize path. You will navigate to this path when starting the OAuth 2.0 dance, for example, `authorize` means the full path is `\http://localhost:8081/authorize`.
+|External Callback Url | You can set this directly to `\http://localhost:8081/callback`.
+|===
+
```
adam-lipka, line 188: I'm not sure where exactly, but we should add a more thorough guide on how to configure and authorize an OAuth flow - what is missing here is that the user has to also configure an HTTP listener, mentioning how the Mule uses the listener for doing the OAuth dance and how the user needs to authorize first before making any requests.
maybe we should add this guide post-release and then reuse it across multiple OAuth connectors?
CC @Hazarth

Hazarth: I agree. Really we should have an entire page somewhere in the Mule Docs just about OAuth and setting up OAuth for CloudHub and connectors should just Link to that page using a _Learn More_ link

OAuth setup is an entire big thing and it keeps evolving, for example we don't even support all the OAuth type configurations which should definitely be documented somewhere. The docs should probably also be written by someone with plenty of experience with OAuth and only then reviewed and finished up by a technical writer 🤔

Could we get an entire page dedicated just for OAuth somewhere @sathya0 ? Or does such a page already exists? If yes then it should be linked here, but if not, we should allocate time to create one, because we found out that this is a very confusing issue for customers that don't have experience with OAuth

CC @adam-lipka

adam-lipka (approved):

tasmoore (approved):

veevavault/1.4/modules/ROOT/pages/veevavault-connector-studio.adoc
sathya0: hey @Hazarth and @adam-lipka, happy to add this page into our docs. I agree we can link to this page from the Authorize Path parameter. This info might be documented somewhere in our Mule docs. Do these pages cover this info? 

https://docs.mulesoft.com/mule-sdk/1.1/oauth-configuring 
https://docs.mulesoft.com/mule-sdk/1.1/oauth-dance
https://docs.mulesoft.com/mule-sdk/1.1/authorization-code

adam-lipka: @sathya0 This one https://docs.mulesoft.com/mule-sdk/1.1/oauth-dance covers it, but there are a few issues IMO: The first is that the doc page is hidden under the Mule SDK section and which is targeted for connector developers, not for connector users, so they have little chance of finding that page. 
The second problem is that it seems that some customers need to have more explicit explanation of what to do - probably even something like:
- hit the xxxx URL in a browser
- login in a login window
- now the connector is authenticated

Then it lacks the coverage of ClougHub x OAuth configuration, which comprises multiple things you have to set up and would be nice to have in under one page.

PR #2484, opened 2024-02-07
W-14974659: MQ-cross-region-resiliency-doc-updates

hannanelson:
# Writer's Quality Checklist

- [ ] Run spell checker
- [ ] Check for orphan files
  - Broken images
  - Dead links
  - Formatting issues, such as:
    - Messed up tables
    - Proper indentation
    - Correct header levels
  - Technical SME
  - Product Manager
  - Editor or peer reviewer


anypoint-mq/4.x/modules/ROOT/pages/anypoint-mq-failover.adoc
```diff
@@ -21,6 +21,10 @@ To use the failover feature, do the following:
 * In Anypoint MQ:
 .. Create or modify a queue and xref:mq::mq-failover.adoc#enable-failover[enable failover on the queue].
 
+[NOTE]
+With failover enabled, the app can't consume directly from the fallback queue.
```
GouthamiKondakindi, line 25: app can't publish to fallback queue directly either.. so we should change this to publish or consume

GouthamiKondakindi, line 25: Should we provide some reasoning for why we do this? When failover system property is enabled, MQ Connector will handle processing messages from both the primary and fallback queues and apps will not be allowed to consume from the fallback queue directly to avoid interrupting the processing that the Connector is doing.

anypoint-mq/4.x/modules/ROOT/pages/anypoint-mq-failover.adoc
```diff
@@ -31,13 +35,40 @@ To set this application property in Runtime Manager:
...
-image::mq::mq-set-properties-failover-enabled.png["Properties tab with failover properties specified"]
+image::mq-set-properties-failover-enabled.png["Properties tab with failover system property set to true"]
 
 Alternatively, set the property in *Run Configurations* in Anypoint Studio.
 For more information, see xref:mule-runtime::mule-app-properties-system.adoc[System Properties in Mule Apps].
 
+[[disable-failover-feature]]
+== Disable Failover in Your App
+
+To disable the failover feature in your app, set the `cross.region.failover.feature.enabled` system property to `false`.
```
GouthamiKondakindi, line 48: I'm wondering if this section will cause confusion. When I read this, the first question that comes to my mind is if I can simply disable the feature from the app itself, then why do I need to turn off the feature from the UI.

When customers turn off the feature from the UI, they just need to restart their apps. They don't really need to set this system property to false. 
But leaving this system property as true will prevent them from consuming from the fallback queue directly. So customers need to turn off this system property for any application where they intend to consume from the fallback queue directly.
I think we should also add a screenshot to explain what "consume directly from fallback queue means" - specifying the fallback queue name as the "Destination" for Publish/Consume and as "Queue" for Subscribe

anypoint-mq/4.x/modules/ROOT/pages/anypoint-mq-failover.adoc
hannanelson: Yes, but not in a note, and maybe not here. I think it belongs in the other section...

anypoint-mq/4.x/modules/ROOT/pages/anypoint-mq-failover.adoc
hannanelson: @GouthamiKondakindi - is there any use case, besides consuming directly from the fb queue, for disabling failover on an app?

GouthamiKondakindi: Not really.. that's the main usecase.. 

We added that system property in the app as a safety mechanism.. if customers are running into some issue with the latest code we released, instead of rolling back to a previous version of connector, they can set this system property to false and we built the connector so that none of the cross region fallback code in the connector will be executed when the system property is false. But we don't need to document this :)

hannanelson:
Work is over there at #2487

PR #2544, opened 2024-04-04
on-error-continue

ca-stefan-cordes:
correct xml start tag.


kevintroller:
Created new PR to add this change. https://github.com/mulesoft/docs-connectors/pull/2635

PR #3237, opened 2026-06-10
@W-22933350-mongo-db-system-property-patch-kt

kevintroller:
## Summary

- Documents the new opt-in system property `com.mulesoft.connector.mongo.skipIsoCoercionUnderOperatorKeys` introduced in MongoDB Connector 6.3.13
- Adds a new `== System Properties` section to the MongoDB Connector 6.3 Reference page, following the same pattern as JMS connector reference docs
- Covers the property's default value (`false`), when to use it (queries on String-typed date fields returning empty results), and caveats (JVM-wide scope, restart required, Extended JSON alternative)

## GUS

[W-22933350](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002bxVUzYAM/view)

## Test plan

- [ ] Verify the new `== System Properties` section renders correctly in the published reference page
- [ ] Confirm table columns (Property, Default, Description) display as expected
- [ ] Confirm the NOTE block renders with both caveats

### DIFF
--- a/mongodb/6.3/modules/ROOT/pages/mongodb-connector-reference.adoc
+++ b/mongodb/6.3/modules/ROOT/pages/mongodb-connector-reference.adoc
@@ -1613,6 +1613,23 @@ Configures the truststore for TLS.
 |===
 
 
+== System Properties
+
+You can configure the following system property for the connector:
+
+[%header,cols="55s,15a,30a"]
+|===
+| Property | Default | Description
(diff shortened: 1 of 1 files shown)


IsaacEldridge (dismissed):

tomastoloza (approved):
Approving — technically verified accurate against the connector source and tech spec (W-22435658). Property name `com.mulesoft.connector.mongo.skipIsoCoercionUnderOperatorKeys`, default `false`, the $-operator passthrough behavior, operator examples, JVM-wide/restart caveats, and the Extended JSON alternative all match the implementation. Version targeting (6.3 reference page) is correct for 6.3.13. Optional follow-up (non-blocking): note the 7.0 deprecation (W-22607949) so adopters plan migration.

IsaacEldridge (approved):

PR #3272, opened 2026-08-03
W-23683594--Azure-Service-Bus-Connector-3.8.0-minor-release-sis

issuvarna1:
# Writer's Quality Checklist

- [ ] Run spell checker
- [ ] Check for orphan files
  - Broken images
  - Dead links
  - Formatting issues, such as:
    - Messed up tables
    - Proper indentation
    - Correct header levels
  - Technical SME
  - Product Manager
  - Editor or peer reviewer


nitesh-salesforce:
## Code Review — 10 findings

Reviewed `microsoft-azure-sb/3.8/**` against the 3.7 baseline. The reference doc has systematic structural problems introduced by this PR: the Message Listener parameter table has been contaminated with Send-only fields, and most operation tables now have duplicate rows for the same attribute. Findings are ranked most-severe first.

---

### 1. Message Listener parameter table contaminated with Send-only fields
**File:** `microsoft-azure-sb/3.8/modules/ROOT/pages/azure-service-bus-connector-reference.adoc` (line 473)

The Message Listener params table (lines 443–497) now contains ~24 Send-op-only rows (`Body`, `Message Id`, `Send Correlation Strategy`, `Correlation Id`, `Reply To`, `Reply To Session Id`, `Label`, `Partition Key`, `Time To Live`, `Time To Live Time Unit`, `Scheduled Enqueue Time Utc`, `Properties`) — none of which apply to a source.

**Impact:** A developer configuring `<azure-service-bus-messaging:message-listener>` will attempt to set attributes the XSD does not expose on the source. Builds fail or the attributes are silently ignored.

---

### 2. Duplicate session-id rows on Message Listener
**File:** `azure-service-bus-connector-reference.adoc` (lines 454 and 472)

Both `Message Session Id` (new, line 454) and `Session Id` (existing, line 472) appear in the same table. In 3.7 there was only `Session Id`; the rename was applied by adding the new row without removing the old one.

---

### 3. Duplicate `Name` row in Configurations table
**File:** `azure-service-bus-connector-reference.adoc` (line 24)

The Configurations table has both:
- Line 14: `|Name | String | Name for this configuration...`
- Line 24: `| Name a| String | ID used to reference this configuration...`

The XSD exposes only one `name` attribute on `<azure-service-bus-messaging:config>`; the second row is a spurious addition.

---

### 4. Duplicate `Configuration` / `Config Ref` rows across every operation table
**File:** `azure-service-bus-connector-reference.adoc`

Each operation table now lists both a `Configuration` (String) row and a `Config Ref` (ConfigurationProvider) row, both marked required, both describing the same connector-configuration reference:

| Operation | Rows |
|---|---|
| Send | 351 & 354 |
| Receive | 259 & 262 |
| Receive Batch | 307 & 311 |
| Send batch of messages | 410 & 413 |
| Message Listener | 447 & 468 |

The connector's XSD exposes a single `config-ref` attribute; the doc now says users must set two required config references per operation.

---

### 5. Message Listener output `Type` silently changed from `Binary` to `Any`
**File:** `azure-service-bus-connector-reference.adoc` (line 502)

3.7 documented the listener output type as `Binary`; 3.8 changes it to `Any` with no release note, description change, or upgrade guidance. If this reflects a real behavior change it affects DataWeave inference and downstream typed components; if it is a doc error, the type contract for the flagship 3.8 source is misstated.

---

### 6. `Consume From Dead Letter Queue` description references old field name
**File:** `azure-service-bus-connector-reference.adoc` (line 463)

The description still says *"Cannot be combined with a Session Id"*, but the listener field was renamed to `Message Session Id` (line 454) and there's a separate `Enable Dynamic Sessions` incompatibility clause on line 464. Reader can't tell which of the three the restriction refers to.

---

### 7. `Message Session Id` row lacks incompatibility note
**File:** `azure-service-bus-connector-reference.adoc` (line 454)

Description is just *"Message session ID."* — no mention that it must be omitted when `Enable Dynamic Sessions` is `true` (documented separately on line 464). A user picking the session field first will misconfigure and only see the failure at runtime.

---

### 8. MongoDB text leaked into a new location
**File:** `azure-service-bus-connector-reference.adoc` (line 494)

`Time To Live Time Unit` description reads *"Time unit for the Write Concern Timeout"* — Write Concern is a MongoDB concept. The pre-existing bug on the Send op (line 376) is out of scope, but the same broken text was replicated into the Message Listener table when Send-op fields were duplicated there.

---

### 9. Dynamic-sessions example uses XSD attribute names not documented anywhere else
**File:** `microsoft-azure-sb/3.8/modules/ROOT/pages/azure-service-bus-connector-examples.adoc` (line 561)

The example uses `dynamicSessionEnabled="true"` and `sessionIdleTimeout="PT30S"` on `<azure-service-bus-messaging:message-listener>`. The reference table only lists display names ("Enable Dynamic Sessions", "Session Idle Timeout"), and the nearby listener examples already use inconsistent casing (`sessionID` vs `sessionId`). If the XSD uses `enableDynamicSessions` instead (a common camel-case convention that matches the display name), users copying the example verbatim will hit a schema-validation error at deploy. Please verify against the 3.8 XSD.

---

### 10. `+` continuation attaches warning to the wrong list item
**File:** `microsoft-azure-sb/3.8/modules/ROOT/pages/azure-service-bus-connector-studio.adoc` (line 74)

```
. (Optional) To receive messages... select *Enable Dynamic Sessions*. When...
* Use *Number Of Consumers*...
* Use *Session Idle Timeout*...
+
*Enable Dynamic Sessions* cannot be combined with a fixed *Message Session Id* or with *Consume From Dead Letter Queue*.
```

The `+` continuation attaches the incompatibility warning to the preceding `Session Idle Timeout` bullet, scoping the warning to that sub-bullet in the rendered output rather than to the parent numbered step.

bjoson-mulesoft (approved):

